PR #11, opened 2024-08-08
Updated conda recipe.

MichaelBarnett:
# Description
Update to conda recipe, to pass local linting and build from bioconda-utils tool. ready for testing with Pull Request into bioconda-recipes repo.

changes:
* Add run_exports field to support others referencing commec versions.
* Add noarch python
* Add minimum versioning for all requirements.
* added Sha256 hash for download confirmation
* Added tagged 0.1.2 tar download for commec as stable URL.

### DIFF
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -1,41 +1,48 @@
-{% set version = "0.1.0" %}
+
+{% set name = "commec" %}
+{% set version = "0.1.2" %}
+{% set sha256 = "33e99060dca151cb9d5481f47e02f1b5276b2cdf61804e7093f0cfc32bef1188" %}
 
 package:
-  name: commec
-  version: {{ version }}
+  name: "{{ name }}"
+  version: "{{ version }}"
 
 source:
-  path: ../  # Adjust as necessary
+  url: https://github.com/ibbis-screening/common-mechanism/archive/refs/tags/v{{version}}.tar.gz
+  sha256: {{ sha256 }}
 
 build:
   number: 0
+  noarch: python
   script: "{{ PYTHON }} -m pip install . --no-deps --no-build-isolation --no-cache-dir -vvv"
+  run_exports:
+    # consider using "x.x.x" rather than "x.x", or "x" considering alpha software and liklihood of change.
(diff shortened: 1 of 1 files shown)


conda-recipe/meta.yaml
```diff
@@ -1,41 +1,48 @@
...
-  path: ../  # Adjust as necessary
+  url: https://github.com/ibbis-screening/common-mechanism/archive/refs/tags/v{{version}}.tar.gz
+  sha256: {{ sha256 }}
 
 build:
   number: 0
+  noarch: python
   script: "{{ PYTHON }} -m pip install . --no-deps --no-build-isolation --no-cache-dir -vvv"
+  run_exports:
+    # consider using "x.x.x" rather than "x.x", or "x" considering alpha software and liklihood of change.
```
alexanian, line 19:
```suggestion
    # consider using "x.x.x" rather than "x.x", or "x" considering alpha software and likelihood of change.
```
tiny typo fix

alexanian: I think we should be okay just to pin the minor version; the patches shouldn't change any interfaces, just bugfixes. Do we need to cut a non-prerelease version? I was looking at the conda docs and it [seems like we're okay to directly specify the version](https://docs.conda.io/projects/conda-build/en/latest/resources/variants.html#id1) but I'd appreciate a sanity check: 

image

alexanian (approved):
Tiny typo to commit, but otherwise LGTM

I had to sanity-check that we're okay pinning this version of Python, but I think the only reason we couldn't before was a now-removed pybedtools dependency (https://github.com/ibbis-screening/common-mechanism/commit/b0c162440a76d09e49855462f08242cf662820cd)

alexanian:
Alright, I approved prematurely (sorry for not reviewing more thoroughly last week!)

I ran into an issue when installing the recipe we've added to bioconda, which is that the `.sh` scripts used to run blast and diamond aren't pulled along:

```
FileNotFoundError: [Errno 2] No such file or directory: '/home/ec2-user/.conda/envs/commec/lib/python3.12/site-packages/commec/run_blastx.sh'
```
Let's discuss options for how best to include those in the package; could be that we wrap the commands in python, there are other options, too!

Issue #10, opened 2024-08-05
Argument discrepancies between run_pipeline.sh and commec wiki


brmagalis:
Line 21 of run_pipeline.sh still shows -p argument as reserved for number of processes.

alexanian:
I think we should be using `commec screen` as opposed to `run_pipeline` by default, but the two should be in agreement with each other as long as we're maintaining both.

Investigating this caused me to notice that `processes` is not really an accurate way to describe how that argument is used; renamed to `jobs` and set up a new default in https://github.com/ibbis-screening/common-mechanism/pull/14

Will update the wiki once we release a new version.

PR #14, opened 2024-08-16
Provide Diamond with --jobs, not "--processes"

alexanian:
# Description
When investigating the inconsistent `--processes` tag, I realised that `run_diamond.sh` was not written in a robust way, and "processes" is not a good description of its function.

Specifically, in this command (pulled out the parts not relevant to parallelization):

```bash
ls ${DB_PATH}/nr*.dmnd | parallel -j ${PROCESSES}
diamond blastx --quiet -d ${DB_PATH}/nr.{%}.dmnd \
--threads ${THREADS} -o ${OUTPUT}.{%}.tsv
```

* `PROCESSES` is not really "processes", it's the number of jobs we tell parallel to run at once
* we use the `{%}` replacement string, which is the job number, to decide which database we use in `--d`... I think they way we've set this up we need to literally always be counting from 1-6. (which is what `screen.py` was doing, but not `run_pipeline.sh`). Based on some local testing of parallel, if we were running with only 2 processes, then the outputs would get (I think? not sure what diamond does with pre-existing output files) overwritten

I have changed this to set the number of jobs to run in parallel based on the threads and CPUs (unless a number is supplied) and made some other parts of run_diamond a bit more robust, then propagated those changes.

**Issues**:
Addresses #10, though we'll need to change the wiki once this is released

## Type of change
- [x] Bug fix (non-breaking change which fixes an issue)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected) _(well, it's a breaking change to run_pipeline, but not to commec screen, which is the preferred interface)_

### DIFF
--- a/commec/run_blastx.sh
+++ b/commec/run_blastx.sh
@@ -10,7 +10,7 @@ set -eu
 DB=""
 QUERY=""
 OUTPUT="out" 
-THREADS=1           #threads per process
+THREADS=1
 FURTHEROPT=""
 
 #Get options from user
--- a/commec/run_diamond.sh
+++ b/commec/run_diamond.sh
@@ -1,89 +1,89 @@
-#! /usr/bin/env bash
+#!/usr/bin/env bash
 # Copyright (c) 2021-2024 International Biosecurity and Biosafety Initiative for Science
 #####################################################################
-#run_diamond.sh runs DIAMOND against a specified NCBI nr database. 
+# run_diamond.sh runs DIAMOND against a specified NCBI nr database.
+# DIAMOND citation: 
 #   Buchfink B, Reuter K, Drost HG, "Sensitive protein alignments at 
 #   tree-of-life scale using DIAMOND", 
 #   Nature Methods 18, 366–368 (2021). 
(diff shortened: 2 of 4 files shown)


MichaelBarnett (approved):
If Threads is not set by the user, then it takes the default value of 1. Requested threads is calculated as 0, from int(1/6)... as processes is the magic number, 6. 
Therefore 0 is the -t arg for run_diamond.sh within screen.py->screen_proteins()

I am unsure how diamond would handle a 0 being passed as a -t, but regardless this appears to be an issue with the original code as well.

Nothing ringing alarm bells otherwise. 
Approved as the above defaults issue already is present.

alexanian:
Thanks for raising that, I hadn't caught the possible weird problem with integer division. I think (more or less by accident) this will be fine in this case, because GNU parallel interprets `--jobs 0` as "run as many jobs in parallel as possible" according to [the docs](https://www.gnu.org/software/parallel/parallel_tutorial.html).

In the case where threads is 1, we aren't concerned about needing to manage the number of jobs because we're not multithreeading, so `-j 0 --use-cpus-instead-of-cores` is the behaviour you want.

PR #15, opened 2024-09-04
Input manager refactor

MichaelBarnett:
# Description

PR makes changes to the handling of inputs, refactoring the code-base, and changing all database calls to pythonic versions, inline with upcoming conda recipe changes. It prepares commec for the json output update.

Tested in fast mode, and unit tests, requires testing with proper blastn / blastx / and diamond databases on AWS before merging.

### Restructured code base into modular file system
* **config**: Contains input and output related parameters and configurations. Including the configuration of default settings for each database being called.
* **screeners**: contains the logic for checking the outputs of database calls. Currently in a state where further refactoring is required to take some logic out of screen.py, and put into the check_*.py files.
* **tests**: Contains all relevant unit test information.
* **tools**: Contains handlers for all databases used, as well as generic tools used for general database parsing, such as BlastTools.
* **utils**: Currently contains orphaned scripts, and static library of file tools, all of which is planned to be folded into other parts of the code base.
* screen.py
* split.py
* flag.py
* cli.py

### Major changes
- entry point renamed to cli.py to reflect its function.
- environment.yml updated to ensure that diamond 2.1.9 is used for diamond based searches.
- Added database handlers, with deal with managing the directory, file, input and output, and command line calls of all database calls for hmm, cmscan, diamond, blastx, blastn etc.
- Added unit tests for databases, and IO parameters
- Added IO parameter class to hold and parse all input information, and serve as general hub of input and parsed user settings/configurations.
- Updated screen_database.py to use the database handler tools, and defaults for how databases are called.
- Default user imports are now set within the ScreenInputParameters dataclass, and are used as defaults when declaring arguments in screen.py to remove double hard-coding.

**Issues**:

## Type of change
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Refactor (non-breaking change to code or documentation which should not change functionality)


commec/config/io_parameters.py
```diff
@@ -0,0 +1,209 @@
...
+    def __init__(self, args : argparse.ArgumentParser):
+        # Ensure required arguments are present
+        required_args = ['fasta_file', 'threads', 'protein_search_tool', 'fast_mode',
+                         'skip_nt_search', 'cleanup', 'output_prefix', 'database_dir', 'jobs']
+        for arg in required_args:
+            if not hasattr(args, arg):
+                raise ValueError(f"Missing required argument: {arg}")
+
+        # Inputs
+        self.inputs : ScreenInputParameters = ScreenInputParameters(
```
alexanian, line 95: Not sure about this naming, since the query and database dir are also inputs

commec/config/io_parameters.py
```diff
@@ -0,0 +1,209 @@
...
+
+        # TODO: Think about whether logs belong in here, or externally.
+
+        # Outputs
+        self.output_prefix = self.get_output_prefix(args.fasta_file, args.output_prefix)
+        self.output_screen_file = f"{self.output_prefix}.screen"
+        self.tmp_log = f"{self.output_prefix}.log.tmp"
+
+        #Query
+        self.query : Query = Query(args.fasta_file, f"{self.output_prefix}.transeq.faa")
```
alexanian, line 112: Can `Query` instead take an input and an output prefix in a constructor? I feel like only `Query` should need to know that we've chosen to call our translated file `f"{self.output_prefix}.transeq.faa")`

commec/config/screen_databases.py
```diff
@@ -0,0 +1,83 @@
...
+
+        if params.should_do_protein_screening:
+            if params.inputs.search_tool == "blastx":
+                self.protein_db = BlastXDataBase(
+                    os.path.join(params.db_dir, "nr_blast"),
+                    os.path.join(params.db_dir, "nr_blast/nr"),
+                    input_file = params.query.fasta_aa_filepath,
+                    out_file = f"{params.output_prefix}.nr.blastx"
+                )
+            elif params.inputs.search_tool == "nr.dmnd":
```
alexanian, line 47: I think we should standardize on `blastx` and `diamond` here (i.e. the sh command used).

commec/config/screen_databases.py
```diff
@@ -0,0 +1,83 @@
...
+        if params.should_do_biorisk_screening:
+            self.biorisk_db = HMMDataBase(
+                os.path.join(params.db_dir, "biorisk_db"),
+                os.path.join(params.db_dir, "biorisk_db/biorisk.hmm"),
+                params.query.fasta_aa_filepath,
+                f"{params.output_prefix}.biorisk.hmmscan",
+            )
+
+        if params.should_do_protein_screening:
+            if params.inputs.search_tool == "blastx":
```
alexanian, line 40: This further makes me think  `inputs` should be called `run_config` or `screen_config`

commec/config/screen_databases.py
```diff
@@ -0,0 +1,83 @@
...
+    """
+    Consolidation and initialisation of all databases needed for the commec workflow.
+    This could be maybe split into a Biorisk_handler, Taxonomy handler, and Benign handler
+    class objects respectively.
+    """
+    def __init__(self, params : ScreenIOParameters):
+        # Biorisk
+        self.biorisk_db : HMMDataBase = None
+        # Taxonomy
+        self.protein_db : BlastXDataBase = None
```
alexanian, line 24: Is this going to be an issue, given that we sometimes set this as a `DiamondDatabase`?

commec/config/screen_databases.py
```diff
@@ -0,0 +1,83 @@
...
+                    os.path.join(params.db_dir, "nr_blast"),
+                    os.path.join(params.db_dir, "nr_blast/nr"),
+                    input_file = params.query.fasta_aa_filepath,
+                    out_file = f"{params.output_prefix}.nr.blastx"
+                )
+            elif params.inputs.search_tool == "nr.dmnd":
+                self.protein_db = DiamondDataBase(
+                    os.path.join(params.db_dir, "nr_dmnd"),
+                    os.path.join(params.db_dir, "nr_dmnd/nr.dmnd"),
+                    input_file = params.query.fasta_aa_filepath,
```
alexanian, line 51: I like that we not have a standard input_file, out_file interface for these, and that this class is essentially the gatherer of magic strings related to the various tools we used in our pipeline!

alexanian (commented):
Review not complete ― I've only really gone into the `config`. Mostly: feeling good about the progress here, and will continue to review after I've checked off a few more things for today.


I'm feeling like `Query` has enough not-really-about-configuration logic in it that we should pull it out of `io_parameters` (maybe just in the base directory)? I think the `tools` classes should be renamed to `ToolHandler` instead of `ToolDatabase`, and `screen_databases` to `screen_tools`, but I do really like that we have a standard interface for them, and we have a config class that accepts the parameters and then sets up all the magic strings.

commec/config/io_parameters.py
```diff
@@ -0,0 +1,209 @@
...
+    """
+    Namespace for optional input parameters for screening; provided by parsing arguments. 
+    All have default values, which are set here, and are reflected into argparse.
+    """
+    threads: int = 1
+    search_tool: str = "blastx"
+    in_fast_mode: bool = False
+    skip_nt_search: bool = False
+    do_cleanup: bool = False
+    diamond_jobs : int = None
```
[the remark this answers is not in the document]
MichaelBarnett: Hmm, not sure, Python has no typing, and in this case, None is the default for when an int isn't set. Not sure if we gain anything from using optional. 
Alternatively we could pass this as -1, which encodes to not using diamond jobs?

commec/config/io_parameters.py
```diff
@@ -0,0 +1,209 @@
...
+Defines the `Input and Output Screen Parameters` class, and associated dataclasses.
+"""
+import os
+import glob
+import argparse
+from dataclasses import dataclass
+import subprocess
+
+@dataclass
+class ScreenInputParameters:
```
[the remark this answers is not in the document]
MichaelBarnett: Sure, as a member variable of Screen, it is labelled as inputs, so having Inputs : ScreenConfig = la de da, makes sense to me. I do like the word inputs, as it clearly labels these as input configurations from the user - rather than derived configurations that are hard coded or calculated from other things.

commec/config/io_parameters.py
MichaelBarnett: Good catch, will do.

commec/config/screen_databases.py
MichaelBarnett: sounds good, I didn't like nr.dmnd, but simply used as it was previously the keyword used and wouldn't want to break anyones scripts.
How about I code for either those or new standards for a short time, until one day we can remove the older nr.dmnd keyword.

commec/config/screen_databases.py
MichaelBarnett: Hmmm I do like the use of input somewhere, as these are specifically those configurations able to be input from the user... Not wedded to the nomenclature, I suppose config works in a similar vein.

commec/config/screen_databases.py
MichaelBarnett: Typing is just a suggestion in Python, and not enforced - but yes - initially I was just going to make all of these the abstract base class of DatabaseHandler, such that they could be changed however, but then figured using derived classes made sense for the commec specific use case.

This shouldn't cause any bugs or issues, but perhaps we simply change to the base class.

commec/config/screen_databases.py
MichaelBarnett: renamed to config in latest version.

commec/config/screen_databases.py
MichaelBarnett: removed type suggestion, as its set appropriately further on in __init__

MichaelBarnett:
Commit made to address previous comments and issues.

commec/config/io_parameters.py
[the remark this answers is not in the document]
alexanian: Looked it up in PEP 484: https://peps.python.org/pep-0484/

It seems like Optional is mostly meant for function annotations?
image

image

So I would say we don't need to worry about this

commec/config/screen_tools.py
```diff
@@ -4,12 +4,13 @@
...
-from commec.tools.blastdmnd_handler import DiamondDataBase
-from commec.tools.cmscan_handler import CmscanDataBase
-from commec.tools.hmm_handler import HMMDataBase
+from commec.tools.blastn_handler import BlastNHandler
+from commec.tools.blastx_handler import BlastXHandler
+from commec.tools.blastdmnd_handler import DiamondHandler
+from commec.tools.cmscan_handler import CmscanHandler
+from commec.tools.hmm_handler import HMMHandler
 
 class CommecDatabases():
```
alexanian, line 15: Then this becomes `ScreenTools`

and we get `self.biorisk_screen_tool`, `self.protein_screen_tool` (or `self.biorisk_tool`), yeah?

commec/screen.py
```diff
@@ -245,7 +245,7 @@ def screen_nucleotides(self):
         """
         # Only screen nucleotides in noncoding regions
         fetch_noncoding_regions(self.databases.protein_db.out_file,
```
alexanian, line 247: hm, I wonder if we want to give the out_file as an output of the `.screen()` method for each tool in future?

commec/screen.py
```diff
@@ -265,7 +265,7 @@ def screen_nucleotides(self):
         logging.debug("\t...checking blastn results")
         check_for_regulated_pathogens(self.databases.nucleotide_db.out_file,
                                        self.params.db_dir,
-                                       str(self.params.inputs.threads))
+                                       str(self.params.config.threads))
 
     def screen_benign(self):
```
alexanian, line 270: While we're editing this, should we move this check to the beginning of the method?

```
        if not os.path.exists(sample_name + ".reg_path_coords.csv"):
            logging.info("\t...no regulated regions to clear\n")
return
```

This is another place where we're using "this file exists" as a proxy for "regulated regions were found", and I think in the (future) JSON output refactor we're going to want to actually check the outcome of that step. Another question we'd want to easily be able to ask the JSON structure ― were any regulated regions identified?

alexanian (commented):
A few comments on the most recent commit, will now start reviewing the other changes!

commec/utils/file_utils.py
```diff
@@ -0,0 +1,75 @@
...
+        usage: check that a file is empty
+        input:
+        - name of file
+        """
+        try:
+            abspath = os.path.abspath(os.path.expanduser(filepath))
+            return os.path.getsize(abspath) == 0
+        except OSError:
+            # If there is an error (including FileNotFoundError) consider it empty
+            return True
```
alexanian, line 30:
```suggestion
def is_empty(filepath: str) -> bool:
    """Check if a file is empty or non-existent."""
    try:
        return os.path.getsize(os.path.abspath(os.path.expanduser(filepath))) == 0
    except OSError:
        # Errors such as FileNotFoundError considered empty
        return True
```

Looking at this, we can simplify; I don't think we need that kind of usage info in the docsting (even though I think I might have written it earlier...)

commec/utils/file_utils.py
```diff
@@ -0,0 +1,75 @@
...
+            with open(filepath, "r", encoding="utf-8") as file:
+                for line in file:
+                    # Strip leading and trailing whitespace and check the first character
+                    if not line.strip().startswith("#"):
+                        # Found a hit!
+                        return True
+            return False
+        except FileNotFoundError:
+            # The file does not exist
+            return False
```
alexanian, line 52:
```suggestion
def has_hits(filepath: str) -> bool:
    """Check if a file has any hits (lines that do not start with '#')."""
    try:
        with open(filepath, "r", encoding="utf-8") as file:
            return any(not line.strip().startswith("#") for line in file)
    except FileNotFoundError:
        return False
```
also easily simplified

commec/utils/file_utils.py
```diff
@@ -0,0 +1,75 @@
...
+        """Raise ArgumentTypeError if `path` is not a file."""
+        if not os.path.isfile(path):
+            raise argparse.ArgumentTypeError(f"{path} is not a valid file")
+        if not os.path.getsize(path) > 0:
+            raise argparse.ArgumentTypeError(f"{path} is an empty file")
+        return path
+
+
+    
+
```
alexanian, line 75:
```suggestion
```

alexanian:
Running on my AWS instance, I got the following output to start off:
```
commec screen ./false-positive-fastas/TN_108.fasta -t 8 -d /home/ec2-user/cm-dbs/ -o ./refactor-false-positives/
 Validating Inputs...
Translate nucleic acid sequences
 Validating Inputs...
Bad database file!
Bad database file!
>> STEP 1: Checking for biorisk genes...
        ...running hmmscan
        ...checking hmmscan results
                 --> Biorisks: no hits detected, PASS

 STEP 1 completed at 2024-09-06 02:39:30
        ...running blastx
```

in a case where I am fairly confident the database file is not bad, since I just used it for other screening. Can we modify this to:

1. Validate databases before translating the input query
2. More descriptively log which database file is bad, and how (e.g. directory doesn't exist? file not found?)
3. Abort the run if a database is bad

commec/screen.py
MichaelBarnett: Hmmm could do it as a qol feature for devs I suppose. However it ties the screen call, and the validity of the output to the call, whereas checking files allows for parallelism if desired as the .screen() ultimate just calls subprocesses.

commec/config/query.py
```diff
@@ -0,0 +1,55 @@
...
+            if result.returncode != 0:
+                command_str = ' '.join(command)
+                raise RuntimeError(
+                    f"subprocess.run of command '{command_str}' encountered error."
+                )
+        except RuntimeError as error:
+            raise RuntimeError("Input FASTA {fasta_to_screen} could not be translated.") from error
+        
+    @staticmethod
+    def get_cleaned_fasta(input_file, out_prefix):
```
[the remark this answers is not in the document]
MichaelBarnett: Yea so it doesn't currently need to be static, no where else currently uses it, however I felt like it was something that could come up to be used elsewhere. I'll change away from static to keep it consistent with its current use.

commec/utils/file_utils.py
```diff
@@ -0,0 +1,75 @@
+#!/usr/bin/env python3
+# Copyright (c) 2021-2024 International Biosecurity and Biosafety Initiative for Science
+"""
+Module for FileTools, containing Static functions useful for dealing with common file parsing tasks.
+"""
+
+import argparse
+import os
+
+class FileTools():
```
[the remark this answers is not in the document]
MichaelBarnett: removed Filetools and BlastTools to also no longer use a class for namespacing only static methods to be more pythonic

commec/utils/file_utils.py
MichaelBarnett: Agree with changes to hashits and is empty, also moved to be part of the tool handlers, which is under the context for where they are only used.

commec/config/io_parameters.py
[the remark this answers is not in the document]
MichaelBarnett: hmmm yea I see your point, the issue is with how we use it - within the DiamondHandler, it is checked for None, and if its an int, it adds itself to the -J CLI for diamond. In this sense it retains its None status, and thus we can never assume it is an int. Perhaps Optional is prudent in this case. I'll make the change.

MichaelBarnett:
Addressed previous comments and suggestions in latest series of commits. See change logs =)

alexanian (commented):
Progress! Some concerns about the current parallelization approach for DIAMOND.

I want to push for us to use something other than `DatabaseHandler` as our label for that abstract base class; each handler under `tools` handles a command used to search a database, but (for example) our `DatabaseVersion` right now is returning the version of the tool used for searching, not the database. Additionally, both BLASTX and DIAMOND are searching the database NCBI NR (albeit in different formats)... so it feels like a confusing pointer to me.

Do you think there's any benefit to using the `abc` module for the base class?

commec/tools/database_handler.py
```diff
@@ -0,0 +1,142 @@
...
+    # End override API
+    def check_input(self):
+        """ 
+        Simply checks for the existance of the input file, This is separated 
+        from database location, as sometimes the input file is generated at a 
+        previous step, and may not exist during DatabaseHandler instantiation.
+        """
+        if os.path.isfile(self.input_file):
+            return True
+        return False
```
alexanian, line 69: We don't currently call this anywhere; I would propose either removing or incorporating into a wrapper that first checks the input, then calls `screen`?

commec/tools/database_handler.py
```diff
@@ -0,0 +1,142 @@
...
+@dataclass
+class DatabaseVersion():
+    """ Container class for outputting version related information from a database."""
+    def __init__(self, input_version : str = "x.x.x", input_date : str = "Null", input_comment :str = ""):
+        self.version_string = input_version
+        self.version_date = input_date
+        self.additional_comment = input_comment
+    version_string : str
+    version_date : str
+    additional_comment : str
```
alexanian, line 21: Does it make sense to split into a tool / command version (e.g. "BLAST+ 2.15") and database version (which for BLAST databases we can pull out using blastdbcmd, but it's not semver, it's a build date https://www.ncbi.nlm.nih.gov/books/NBK569853/).

commec/tools/database_handler.py
```diff
@@ -0,0 +1,142 @@
...
+    def screen(self):
+        """ Virtual function to be called by any child database implementation"""
+        raise NotImplementedError(
+            "This class must override the .screen() to use DatabaseHandler."
+        )
+
+    def get_version_information(self) -> DatabaseVersion:
+        """ Override to call version retrieval information on a database."""
+        new_version_info = DatabaseVersion("", "", "Version retrieval not yet supported for this database")
+        return new_version_info
```
alexanian, line 39: Would it make sense to set the version on `init`, rather than having it as a separate method?

commec/tools/database_handler.py
```diff
@@ -0,0 +1,142 @@
...
+
+    def check_output(self):
+        """ 
+        Simply checks for the existance of the output file, 
+        indicating that the database screening ran. Can be overrided if
+        more complex checks for a particular db is desired.
+        """
+        if os.path.isfile(self.out_file): # Consider adding checks from File_Tools such that empty outputs are invalid?
```
alexanian, line 47: Remove comment?

commec/tools/database_handler.py
```diff
@@ -0,0 +1,142 @@
...
+        structurally ready for appending to a command list of strs.
+        """
+        my_list = []
+        for key, value in self.arguments_dictionary.items():
+            my_list.append(str(key))
+            if isinstance(value, list):
+                my_list.append(" ".join(value))  # Extend the list with all elements in the array
+            else:
+                my_list.append(str(value))  # Append the value directly if it's not a list
+        return my_list
```
alexanian, line 101: Might make sense to implement this as an `@property` called `arguments_list` or `arguments_for_command`?

I think you'll also want to do  `my_list.append(" ".join(map(str, value)))` (i.e. explicitly case to string)

commec/config/io_parameters.py
```diff
@@ -0,0 +1,133 @@
...
+from commec.config.query import Query
+
+@dataclass
+class ScreenConfiguration:
+    """
+    Namespace for optional input parameters for screening; provided by parsing arguments. 
+    All have default values, which are set here, and are reflected into argparse.
+    """
+    threads: int = 1
+    search_tool: str = "blastx"
```
alexanian, line 23: `protein_search_tool` or `aa_search_tool` (since we always use `blastn` for nt)?

commec/tools/blast_tools.py
```diff
@@ -110,6 +94,14 @@ def taxdist(blast, reg_ids, vax_ids, db_path, threads):
     reg = list(map(str, reg_ids[0]))
     vax = list(map(str, vax_ids[0]))
 
+    try:
+        a = t["FullLineage"].str.split(";")[0]
+        b = t["FullLineageTaxIDs"].str.split(";")[0]
+        c = t["FullLineageRanks"].str.split(";")[0]
+    except AttributeError:
+        logging.error("The Blast database used has not returned any Lineage information!")
+        return blast
```
alexanian, line 103: What's this test?

alexanian: From conversation: this prevents an error from being thrown if the BLAST database has no taxonomy information. Having thought about this more, we indeed shouldn't fail silently here, nor with a random AttributeError... what if we do something like:

```
    reg = list(map(str, reg_ids[0]))
    vax = list(map(str, vax_ids[0]))

    # checks which individual lines contain regulated pathogens
    lineage_info = pytaxonkit.lineage(blast["subject tax ids"], data_dir=db_path, threads=threads)
    if not all(col in df.columns for col in ["FullLineage", "FullLineageTaxIDs", "FullLineageRanks"]):
        logging.error(f"ERROR: Lineage information not returned based on NCBI taxonomy in: {db_path}")
        return blast
```

commec/tools/blastdmnd_handler.py
```diff
@@ -0,0 +1,99 @@
...
+                "--frameshift", str(self.frameshift),
+                "--outfmt", self.output_format,
+            ]
+            command.extend(self.output_format_tokens)
+
+            if self.do_range_culling:
+                command.append("--range-culling")
+
+            if self.jobs is not None:
+                command.extend(["-j", str(self.jobs)])
```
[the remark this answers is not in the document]
MichaelBarnett: Yeap the Diamond was very broken. Recent commits have reworked, and should address this.

commec/tools/blastdmnd_handler.py
```diff
@@ -0,0 +1,99 @@
...
+            "qseqid",   "stitle",   "sseqid",   "staxids",
+            "evalue", "bitscore", "pident", "qlen",
+            "qstart", "qend",     "slen",   "sstart",  "send"]
+
+    def screen(self):
+        # Find all files matching the pattern nr*.dmnd in DB_PATH
+        db_files = glob.glob(f"{self.db_directory}/nr*.dmnd")
+
+        if len(db_files) == 0:
+            raise FileNotFoundError(f"Mandatory Diamond database directory {self.db_directory} contains no databases!")
```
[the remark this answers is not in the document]
MichaelBarnett: Hmm technically, this check already occurs, when validate is called by the BlastHandler ABC when it is initialised. This second check could simply be removed.

commec/tools/blastdmnd_handler.py
```diff
@@ -0,0 +1,99 @@
...
+
+            log_i = open(log_i_filename, "w", encoding="utf-8")
+            print(" ".join(command))
+            process = subprocess.Popen(command, stdout=log_i, stderr=subprocess.STDOUT)
+            processes.append(process)
+            logs.append(log_i)
+
+        # Wait for all processes to finish
+        for process in processes:
+            process.wait()
```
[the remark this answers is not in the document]
MichaelBarnett: Switched to the better system using Pools from multiprocessing as discussed.

commec/tools/blastdmnd_handler.py
```diff
@@ -0,0 +1,99 @@
...
+
+        # Logs are concatenated into a global diamond log.
+        with open(self.temp_log_file, 'w', encoding="utf-8") as outlog:
+            for log_f in output_lognames:
+                if os.path.exists(log_f):
+                    with open(log_f, 'r', encoding="utf-8") as infile:
+                        outlog.write(infile.read())
+                    os.remove(log_f)
+
+    def get_version_information(self) -> DatabaseVersion:
```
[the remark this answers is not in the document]
MichaelBarnett: All the get versions are probably closer aligned to the json-output branch than the refactor. Perhaps we just removed them for from this PR now, and put them back later.

commec/tools/blastn_handler.py
```diff
@@ -0,0 +1,45 @@
...
+"""
+
+import subprocess
+from commec.tools.blast_tools import BlastHandler
+from commec.tools.database_handler import DatabaseVersion
+
+class BlastNHandler(BlastHandler):
+    """ A Database handler specifically for use with BlastX files for commec screening. 
+    Allows for full customization of any of the callable blast flags. A better implementation
+    may be achieved using import blast"""
```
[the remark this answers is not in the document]
MichaelBarnett: Sorry, code thoughts creeping their way into comments, I believe this was removed in a recent tidying commit. Yes I was referring to biopython when considering best practices to call CLIs, but since decided to implement the search_handler CLI system we use now.

commec/tools/blast_tools.py
```diff
@@ -398,8 +247,12 @@ def shift_hits_pos_strand(blast):
             blast.loc[j, "q. end"] = end
     return blast
 
-
 def trim_edges(df):
+
+    # Use this enumeration, instead of below, for pylint errors - needs testing first.
```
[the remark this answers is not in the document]
MichaelBarnett: Cleaned __init__, removed the code suggestion until we make the unit test specifically for trim edges.

commec/tools/blastdmnd_handler.py
```diff
@@ -0,0 +1,99 @@
...
+        processes = []
+        logs = []
+        output_files = []
+        output_lognames = []
+        for i, db_file in enumerate(db_files, 1):
+            output_file = f"{self.out_file}.{i}.tsv"
+            log_i_filename : str = self.temp_log_file + "_" + str(i)
+            output_files.append(output_file)
+            output_lognames.append(log_i_filename)
+            #output_log : str = self.temp_log_file + "_" + str(i)
```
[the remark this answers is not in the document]
MichaelBarnett: diamond has been rewritten.

commec/utils/file_utils.py
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env python3
+# Copyright (c) 2021-2024 International Biosecurity and Biosafety Initiative for Science
+"""
+Module for FileTools, containing Static functions useful for dealing with common file parsing tasks.
```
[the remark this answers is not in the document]
MichaelBarnett: Will keep for now as discussed, until we find better way to share these between screen, flag, and split.

commec/tools/blastn_handler.py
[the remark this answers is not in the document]
MichaelBarnett: Slight alterations to the docstrings for blastn , diamond and Blastx to reflect the handlers.

commec/tools/database_handler.py
MichaelBarnett: Removed, as the input file is either the query (checked by io_parameters), or its the checked output of another search handler, and therefore checked as output.

commec/tools/database_handler.py
MichaelBarnett: Yea why not. We can set version as a member variable.

MichaelBarnett: Although will keep the method, override system, as it is neater to call it within super().__init__ that way.

commec/tools/blastdmnd_handler.py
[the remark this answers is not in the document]
MichaelBarnett: Kept them, and updated them all to work as intended.

commec/tools/database_handler.py
MichaelBarnett: Kept it simple with a string for tool and a string for database, and these can be filled however they like - at the moment all databases are programmed to return a short snippet from various cli's which I deemed would be useful.

commec/config/io_parameters.py
MichaelBarnett: Yeap! I'll make the change.

commec/tools/blast_tools.py
MichaelBarnett: Unfortunately the columns are present, just filled with none string objects, so this approach doesn't work.

alexanian: Just discussed a slight rephrase:
```
# Check that the required lineage information is present; same access as below. Can't just check that columns are present, since if db missing the data, they are present with null values
for required_lineage_col in ["FullLineage", "FullLineageTaxIDs", "FullLineageRanks"]:
    t[required_lineage_col].str.split(";")[0]
```

and then probably actually raise an exception?

alexanian (approved):
Long process, but overall feeling good about these changes, and looking forward to more bugfixes, unit tests, and user-facing improvements that we'll be able to build upon this.

PR #18, opened 2024-10-22
Bugfix: filter out deleted and invalid taxonomy ids

alexanian:
## Background

We were sometimes seeing this kind of error:
```
Problem with taxID 2855502 - check that your taxID and protein database are up to date
Traceback (most recent call last):
  File "/home/repository/commec/check_reg_path.py", line 153, in <module>
    main()
  File "/home/repository/commec/check_reg_path.py", line 57, in main
    blast = taxdist(blast, reg_ids, vax_ids, args.db + "/taxonomy/", args.threads)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/repository/commec/utils.py", line 118, in taxdist
    zip(
TypeError: 'float' object is not iterable
	 ERROR: command 'python /home/repository/commec/check_reg_path.py -i ./output/test.nr.blastx -d /home/cm-dbs/ -t 8' failed
```

after investigation, this turned out to be because `pytaxonkit.lineage` will return a column full of mostly `NaN`s if the taxonomy id has been deleted or cannot be found. Now we log a warning about that and drop the hits with outdated taxids.

### Bug fixes
* Filter out deleted and invalid taxonomy ids
* Remove check for `pytaxonkit.lineage` results introduced in refactor (same issue caught by the filter above)

### Refactoring
* Add tests for some of the `blast_tools.py` methods
* Removed unused method in `blast_tools.py` and gave a few others more descriptive names

### DIFF
--- a/commec/screeners/check_benign.py
+++ b/commec/screeners/check_benign.py
@@ -15,15 +15,16 @@ import os
 import sys
 import pandas as pd
 
-from commec.tools.blastn import BlastNHandler # For has_hits.
-from commec.tools.blast_tools import get_top_hits, readblast, _trim_overlapping
+from commec.tools.blastn import BlastNHandler  # For has_hits.
+from commec.tools.blast_tools import get_top_hits, read_blast
 from commec.tools.hmmer import readhmmer
 from commec.tools.cmscan import readcmscan
 
+
 def check_for_benign(query, coords, benign_desc):
-    '''
-    Checks a query against taxonomy 
-    '''
+    """
+    Checks a query against taxonomy
+    """
     cleared = [0] * coords.shape[0]
 
     # PROTEIN HITS
@@ -55,8 +56,12 @@ def check_for_benign(query, coords, benign_desc):
                     # for row in range(htrim.shape[0]):
                     for row in [0]:  # just print the top hit
                         hit = htrim["target name"][row]
-                        hit_msg = (hit + ": " + str(*benign_desc['Description'][benign_desc['ID'] == hit])
-                                   + f" (E-value: {htrim['E-value'][row]:.3g}")
+                        hit_msg = (
+                            hit
+                            + ": "
+                            + str(*benign_desc["Description"][benign_desc["ID"] == hit])
+                            + f" (E-value: {htrim['E-value'][row]:.3g}"
+                        )
                         descriptions.append(hit_msg + "\n")
                     annot_string = "\n".join(str(v) for v in descriptions)
                     logging.info(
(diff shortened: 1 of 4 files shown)


commec/screeners/check_benign.py
```diff
@@ -124,8 +124,7 @@ def check_for_benign(query, coords, benign_desc):
         logging.info("\t...no Synbio sequence hits\n")
     else:
         blastn = readblast(blast)  # synbio parts
-        blastn = trimblast(blastn)
-        blastn = tophits(blastn)
+        blastn = get_top_hits(blastn)
```
MichaelBarnett, line 127: _trim_overlapping is still in import? Unused, delete? - I see its been folded into get_top_hits. Love the use of private _ modifier. I think there's some more functions around that could benefit from that.

commec/screeners/check_reg_path.py
```diff
@@ -15,7 +15,7 @@
 import sys
 import textwrap
 import pandas as pd
-from commec.tools.blast_tools import readblast, trimblast, taxdist, tophits
+from commec.tools.blast_tools import readblast, get_taxonomic_labels, get_top_hits
```
MichaelBarnett, line 18: thank you for renaming tophits and taxdist, perhaps we should quickly change readblast to read_blast snake casing whilst we're here?

commec/tools/blast_tools.py
```diff
@@ -39,120 +42,145 @@ def _validate_db(self):
...
+
+    # Add new columns, which will later be used to classify the hits as regulated or non-regulated
+    blast["regulated"] = False
+    blast["superkingdom"] = ""
+    blast["phylum"] = ""
+    blast["genus"] = ""
+    blast["species"] = ""
+
+    blast = blast[blast[TAXIDS_COL] != TAXID_SYNTHETIC_CONSTRUCTS]
+    blast = blast[blast[TAXIDS_COL] != TAXID_VECTORS]
```
MichaelBarnett, line 128: and there were no more magic numbers!

commec/tools/blast_tools.py
```diff
@@ -331,16 +328,16 @@ def trim_edges(df):
     return df, rerun
 
 
-def tophits(blast2):
+def get_top_hits(blast: pd.DataFrame):
```
MichaelBarnett, line 331: I always wondered why there were so many references to the same blast dataframe (with blast3 = blast2 etc etc) I figured they were mutable, and would just change the same space in memory. Good to see them gone.

MichaelBarnett (approved):
Works on my machine, and looks good to me! Minor comments on one or two lines, namely erroneous import in benign, and readblast -> read_blast, but will approve and you can make them if you wish, as they are not big issues.
Overall a vast improvement compared to what was there!

commec/tools/blast_tools.py
alexanian: livin' the dream lol

commec/tools/blast_tools.py
alexanian: My guess is that this was just a refactor of an earlier, more pipeline-y script (where progressing through different processing steps as blast1 -> blast2 -> blast2 improved readability) but I'm not sure. I don't think there should be a memory advantage to doing it that way, but I'm not sure

PR #20, opened 2024-10-30
Add --force to Commec Screen CLI

MichaelBarnett:
Added --force, -F, feature to the Commec Screen CLI.

## Background

When re-running Commec Screen, pre-existing outputs caused RunTimeErrors for the pre-existing .screen file. Furthermore, there was little to no control on re-running some database searches whilst skipping others.

## Changes

* Added --force or -F to Commec Screen CLI.
* Databases try to re-use existing output data by default (only applies if .screen file doesn't exist)
* Some minor logging to inform the user that the database outputs are being re-used.
* Search Handler base class now has the private _screen, and screen wrapper, to facilitate the logic of passing force bool.
* io_parameters.py no longer raises an exception when .screen is present unexpectedly, and instead cleaning quits.

### New features
With the addition of a Force parameter, databases can by default check if their outputs already exist, and use them instead. The .screen file can also be over-written with --force. This allows the following highlight example use-cases:
* Rerun Screen entirely - use --force.
* Rerun Screen, whilst not re-running any particular database - delete .screen output, keep any database output.
* Accidentally overwrite any outputs from a previous run? - not possible (without --force) due to pre-existing .screen file check. System exits gracefully.

### Refactoring
* Derived search_handlers now override _screen() private function, rather than screen(force : bool).

### DIFF
--- a/commec/config/io_parameters.py
+++ b/commec/config/io_parameters.py
@@ -31,6 +31,7 @@ class ScreenConfig:
     do_cleanup: bool = False
     diamond_jobs: Optional[int] = None
     force: bool = False
+    resume: bool = False
 
 class ScreenIOParameters:
     """
@@ -46,6 +47,7 @@ class ScreenIOParameters:
             args.cleanup,
             args.diamond_jobs,
             args.force,
+            args.resume,
         )
 
         # Outputs
@@ -60,9 +62,15 @@ class ScreenIOParameters:
         self.db_dir = args.database_dir
 
         # Check whether a .screen output file already exists.
-        if os.path.exists(self.output_screen_file) and not self.config.force:
-            print(f"Screen output {self.output_screen_file} already exists. Aborting.")
+        # Note : The only operational difference between force and resume, is that resume is not passed into the search handlers.
+        if os.path.exists(self.output_screen_file) and not (self.config.force or self.config.resume):
+            print(f"Screen output {self.output_screen_file} already exists. \n"
+                  "Either use a different output location, or use --force, --resume to override. "
+                  "\nAborting Screen.")
             sys.exit(1)
+
+        if self.config.force and self.config.resume:
+            print("Warning: Using --force, overrides --resume behaviour.")
 
     def setup(self) -> bool:
         """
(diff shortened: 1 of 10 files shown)


alexanian:
I think we should pop `force` into the initialization of the `searchhandler` classes, but otherwise LGTM

commec/tools/search_handler.py
```diff
@@ -55,8 +79,21 @@ def temp_log_file(self):
...
+        """
+        Wrapper for _search, to ensure that it is only called if 
+         - The output doesn't already exist,
+         - If force is enabled.
+        """
+        if not self.force and self.check_output():
+            logging.info("%s expected output data already exists, "
+                         "will use existing data found in:\n%s",
+                         self.__class__.__name__, self.out_file)
+            return
```
alexanian, line 92: I think I support splitting this into `--resume` and `--force` (as discussed on our call)

MichaelBarnett:
Updated to also take --resume, instead of manually deleting the .screen file, and running without --force. You can simply used --resume. This behaviour is overridden by --force. (If both are set, a warning is displayed)

commec/config/io_parameters.py
```diff
@@ -59,10 +62,15 @@ def __init__(self, args: argparse.ArgumentParser):
...
-                f"Screen output {self.output_screen_file} already exists. Aborting."
-            )
+        # Note : The only operational difference between force and resume, is that resume is not passed into the search handlers.
+        if os.path.exists(self.output_screen_file) and not (self.config.force or self.config.resume):
+            print(f"Screen output {self.output_screen_file} already exists. \n"
+                  "Either use a different output location, or use --force, --resume to override. "
+                  "\nAborting Screen.")
+            sys.exit(1)
+
+        if self.config.force and self.config.resume:
```
alexanian, line 72: Small change we discussed yesterday is switching this over to using the [`mutually_exclusive_group`](https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.add_mutually_exclusive_group) in `argparse`. Happy to approve after that!

MichaelBarnett: Added, ready to go! =D

alexanian: Removed that warning in 505564f8e9786888121b29dee493f857106dd46d since the mutually_exclusive_group should prevent reaching that state

alexanian (approved):

PR #22, opened 2024-11-08
Bugfix for Benign Search

MichaelBarnett:
benign search was running when the reg_path_coords.csv existing, but was empty, leading to some erroneous outputs to the screen file.

## Changes
Added the check for the shape of the imported csv, checking if it is 0, and returning early if it is. As was previous behaviour pre-refactor.


alexanian (approved):
Thanks for catching this!

PR #24, opened 2024-12-11
JSON and HTML Output

MichaelBarnett:
## Background
Updates the output format potential for Commec, but centralising output Screen Data into a structure of Dataclasses, which are ultimately output as a JSON compatible format, and HTML visualisation.

## Changes
- Overhauls internal state by introducing ScreenData, a hierarchy of dataclasses, which stores the current information pertaining to a screen - which includes multiple Query handling.
- Includes an HTML output to visualise the outputs of Commec Screen.
- Extremely verbose queries which hit thousands of identical hits and regions, now collapse into more - human enjoyable formats.
- Much more control on the logic of how each step affects the overall Commec outcome.

### Bug fixes
- Fixed a bug relating to yaml --resume behaviour not correctly updating the existing db_dir value from yaml, when not overriding from cli.

### New features
- New HTML output file.
- New JSON output file.

### Breaking changes 
- Should be backwards compatible with main.

- Note, we still have issues in indexing for AA/Nucleotide/HMMSCAN indexing at the query bp level. Also doesn't yet address pre-existing pytaxonkit failure issue.
- When an exception is raised, an HTML and JSON will still be attempted to be written, which may create verbose error messages.
- Increases Commec time by a factor of 2 for pythonic check_X components, due to the duplicated logic. Plenty of room for improvement in speed here.


commec/screen.py
```diff
@@ -163,21 +176,38 @@ def add_args(parser: argparse.ArgumentParser) -> argparse.ArgumentParser:
...
-        self.params: ScreenIOParameters = None
-        self.database_tools: ScreenTools = None
-        self.scripts_dir: str = os.path.dirname(__file__)  # Legacy.
+        self.params : ScreenIOParameters = None
+        self.database_tools : ScreenTools = None
+        self.scripts_dir : str = os.path.dirname(__file__) # Legacy.
+        self.screen_data : ScreenData = ScreenData()
+        self.start_time = time.time()
+
+    def __del__(self):
```
alexanian, line 192: citation: talking to an LLM, [this stackoverflow](https://stackoverflow.com/questions/40536821/python-enter-exit-vs-init-or-new-del), but I think we might want to use `__exit__` and define a `_cleanup` method that is called from within it, rather than using `__del__` (which is about garbage collection)?

alexanian (changes requested):
not finished, but here's a partial review to get started looking at

commec/config/json_io.py
```diff
@@ -0,0 +1,453 @@
...
+import os
+from dataclasses import dataclass, asdict, fields, field, is_dataclass
+from typing import Dict, Type, get_origin, Any, get_args, List, Iterator, Tuple
+from enum import StrEnum
+from commec.tools.search_handler import SearchToolVersion
+
+# Seperate versioning for the output JSON.
+JSON_COMMEC_FORMAT_VERSION = "0.1"
+
+def guess_domain(search_string : str) -> str:
```
alexanian, line 36: We only call this from `check_biorisk`, should we put it in there instead?

commec/config/json_io.py
```diff
@@ -0,0 +1,453 @@
...
+    search_token = search_string.lower()
+    if contains(search_token, ["vir", "capsid", "RNA Polymerase"]):
+        return "Virus"
+    if contains(search_token, ["cillus","bact","coccus","phila","ella","cocci","coli"]):
+        return "Bacteria"
+    if contains(search_token, ["eukary","nucleus","sona","odium","myces"]):
+        return "Eukaryote"
+    return "not assigned"
+
+class CommecScreenStep(StrEnum):
```
alexanian, line 58: Restructuring proposal (related to the `config` --> `io`) that I proposed before; I think we should separate out the enums / bits holding the actual progress of screening from the JSON serialization logic.

What if we did `screeners` --> `screening` and moved some of these data classes in there?

commec/config/json_io.py
```diff
@@ -0,0 +1,453 @@
...
+            self.commec_recommendation = CommecRecommendation.WARN
+            return
+        # At the moment we only get here when biorisk screen is just run.
+        # this will set the global recommend to pass or error, after biorisk is done.
+        # We will need earlier checks to maintain the error status of future steps however.
+        if self.commec_recommendation == CommecRecommendation.NULL:
+            self.commec_recommendation = self.biorisk_screen
+
+@dataclass
+class QueryData:
```
alexanian, line 223: Maybe move this and `ScreenData` into something like `screening/results.py` and then call them something like `QueryRun` and `ScreenRun`? (with just `Data` it's not obvious to me that these are output vs. input data containers)

commec/screen.py
MichaelBarnett: I came across this exact same stackoverflow thread when writing this code, and actually did try using __exit__ and using _with_ context, however after testing with exceptions / and forcing stops, I got the most consistent successful calling of the delete logic with the __del__ override - which was surprising as it went against somewhat what was being said, but makes sense from a memory perspective, especially as the screen instances has a very clear life-time and scope, and there is only ever a single reference to it.
 I can try those tests again to see if we can get the same or more expected behaviour with __exit__.

commec/config/json_io.py
```diff
@@ -0,0 +1,453 @@
...
+    '''Container to hold data related to the Query
+      used in a commec database screen'''
+    query : str = ""
+    length : int = 0
+    sequence : str = ""
+    recommendation : CommecRecommendationContainer = field(default_factory=CommecRecommendationContainer)
+    # TODO: Remove summary statistics for first PR?
+    summary_info : CommecSummaryStatistics = field(default_factory=CommecSummaryStatistics)
```
alexanian, line 231: I think we should remove this here, yeah

commec/config/json_io.py
```diff
@@ -0,0 +1,453 @@
...
+    hits : list[HitDescription] = field(default_factory=list)
+    # non_coding_regions : list[list[int]] = field(default_factory=list)
```
alexanian, line 233:
```suggestion
```
Assuming we don't want this here?

commec/config/json_io.py
```diff
@@ -0,0 +1,453 @@
...
+                    new_region.query_end == existing_region.query_end):
+                    is_unique_region = False
+            if is_unique_region:
+                is_updated = True
+                existing_hit.ranges.append(new_region)
+
+        if is_updated:
+            return True
+
+        return False
```
alexanian, line 268:
```suggestion
        return hits_is_updated
```

alexanian (commented):
running slowly through...

commec/screen.py
```diff
@@ -203,7 +233,30 @@ def setup(self, args: argparse.ArgumentParser):
...
+        if self.params.should_do_protein_screening:
+            self.screen_data.commec_info.protein_database_info = self.database_tools.regulated_protein.get_version_information()
+
+        if self.params.should_do_nucleotide_screening:
+            self.screen_data.commec_info.nucleotide_database_info = self.database_tools.regulated_nt.get_version_information()
+
+        if self.params.should_do_benign_screening:
+            self.screen_data.commec_info.benign_protein_database_info = self.database_tools.benign_hmm.get_version_information()
+            self.screen_data.commec_info.benign_rna_database_info = self.database_tools.benign_blastn.get_version_information()
+            self.screen_data.commec_info.benign_synbio_database_info = self.database_tools.benign_cmscan.get_version_information()
```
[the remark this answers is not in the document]
MichaelBarnett: hmm despite nesting being so fun, I see you point. I did a small work-around with a local variable assignment, for readibilities sake. But perhaps that too is not dealing with the original problem. I would rather have obtusely long nested paths, rather than compromise by shorthanding individual variable names however.

commec/config/json_io.py
MichaelBarnett: not a bad idea really, especially as we hope to be rid of it with updated biorisk databases in the future.

commec/config/json_io.py
MichaelBarnett: At the very least a rename is in order as discussed, and perhaps removing any erroneous info we don't wish to be reported on. As you know this has the single purpose of an output representation container for a single query. I like QueryResult and ScreenResult. I suppose I've used data in the past as it aligns with 'State'.

alexanian:
based on our discussion last week, a proposal for reorganizing modules (that does not yet separate out the state-storage vs. output-handling parts of json_io) : https://github.com/ibbis-bio/common-mechanism/commit/db30b3ea9ce02241c30806a5cd26ba49fb3d9887

alexanian:
A few requests from more end-to-end testing:

* Add some guard clauses in `__del__` (specifically: check if screen_io is `None`, since if so the run failed on initialization and the JSON/HTML output should not be produced; right now this leads to an error) 
* Make sure tracebacks are captured in the `log` or `screen` file; should be easy to replicate by passing in an invalid database directory
* We should be referencing `template.html` as a package resource, not a filepath, should be doing a `importlib.resources.files("commec")` call rather than `template = Template(filename="commec/utils/template.html")`
* Should not skip the benign search in the case of the following:
```
            "recommendation": {
                "screen_status": "Flag",
                "biorisk_status": "Warning",
                "protein_taxonomy_status": "Flag",
                "nucleotide_taxonomy_status": "Pass",
                "benign_status": "Skip"
            },
```
* Also, we currently still produce a `.screen` file, but it seems almost exactly the same as the `log` file (except the log contains a dump of the input FASTA; probably not needed?) and is not accurate. For example, for the run above, the `.screen` file was:
```
>> STEP 1: Checking for biorisk genes...
		 --> Biorisks: Regulated genes not found, PASS

 STEP 1 completed at 2025-02-12 21:03:25
 STEP 2 completed at 2025-02-12 21:40:18
 >> STEP 3: Checking regulated pathogen nucleotides...
Checking protein hits in: /mnt/data/home/ec2-user/analysis/functional-test-json/2025-02-11/BBa_I766605-YopH-EE.nr.blastx
Protein hits found for BBa_I766605_YopH-EE_under, fetching nt regions not covered by a 90% ID hit or better
Identified noncoding regions [1414-1641]
		 --> no top hit exclusive to a regulated pathogen: PASS

	...no regulated hits

 STEP 3 completed at 2025-02-12 22:36:45
>> STEP 4: Checking any pathogen regions for benign components...
	...no regulated regions to clear

>> STEP 4 completed at 2025-02-12 22:36:45
>> COMPLETED AT 2025-02-12 22:36:45
```
which doesn't match the JSON.

PR #26, opened 2025-01-14
Hmmer indexing fix

MichaelBarnett:
## Background
Turns out that the AA centric outputs of HMMER from Biorisk were not within the same coordinate space of the NT centric query level coordinates. This leads to several issues, including the misattribution of benign clearing etc to regions that are not even true biorisks. This PR fixes the indexing issue for HMMER Biorisk output, by correctly converting the 6 frames of AA translations, into the nucleotide coordinate space of the input queries. 
It also addresses the behaviour of trimhmmer, which has been renamed, and the functionality refined to only trim where both the hit is fully encapsulated by another, and for when that hits score is less than the encapsulating hit. As well as bug-fixes for multi-query situations.

## Changes
* HMMER import now adds a frame column, with the integer of the frame, derived from the query name_'X'
* Added function to add q. start and q. end columns to a hmmer output calculated from 'frame', ali from and to, and qlen. (Note: it is known that qlen is not the true-ly correct value here, and may be off by 2 or 1 nucleotides depending on frame.)
* trimhmmer now replaced with remove_overlaps, which fixes issues with multi-query FASTAs, which were not being trimmed in that context, as well as appropriately doesn't trim small overlapping queries. Also those queries with high scores, are not trimmed by those larger encapsulating hits with lower scores.

### Bug fixes
* Fixes outputs of HMMER to contain and report biorisks in query coordinate space (nucleotides)
* Fixes TRIM HMMER to correctly work on multi-query FASTAs.


MichaelBarnett:
NOTE: There is currently an assumption built into this solution that the nucleotide counts of the original frame are infact a multiple of 3. And there may be off-by-one, or two, errors in cases where there are non nucleotides%3 == 0 examples for the 1st reading frame. This will need to be fixed in tandem with updating the hmmer with original NT length depending on query name.

alexanian:
Per the discussion @MichaelBarnett and I just had, let's merge this into https://github.com/ibbis-bio/common-mechanism/tree/json-output,, then we'll also plan to merge https://github.com/ibbis-bio/common-mechanism/tree/remove-emboss-dep into that branch, then should have something clean with few indexing errors.

alexanian (approved):
Changed target to json-output, think we're okay to merge

Issue #28, opened 2025-01-21
Docker container?


mcrone:
Thanks for the great contributions! Along with the conda environment, it would be great to have a minimal docker container with everything needed. Is this something that could be readily added (by just creating a base container with the conda install based on the recipe?). Also happy to create a dockerfile myself, but would be great if there was just a standard image from IBBIS.

mcrone:
Thanks for the new docker container version!

I am being a little cheeky here, but it would be very useful to have a self-contained container that already has the databases needed to run the --fasta version of the screening. I plan to use this as a part of a nextflow pipeline and the workflow to have a database and then mount that is more complicated than I would like. Alternatively, is there a possibility of sharing the dockerfile that quay uses to build the image?

mcrone:
This was me being too lazy, here is the Dockerfile code for a minimal container with the databases needed for running fast screening found within the container itself, rather than needing to link to an external database.

```
FROM quay.io/biocontainers/commec:0.3.1--pyhdfd78af_0

RUN wget -O commec-dbs.zip https://f005.backblazeb2.com/file/common-mechanism-dbs/common-mechanism-dbs.zip \
&& unzip 'commec-dbs.zip' -d /commec-dbs && rm \
commec-dbs.zip || true;
```

PR #31, opened 2025-01-30
Remove EMBOSS transeq use, fix hmmer indices

alexanian:
## Background

The  `biocontainer` auto-built from the `commec` bioconda packages [currently has some gnarly security warnings](https://quay.io/repository/biocontainers/commec) (related to https://github.com/bioconda/bioconda-recipes/issues/46980#issuecomment-2604135864 ). This removes that dependency by writing our own `translate` method.

Also updates #26 (I have marked this as a draft, since that one should be merged first) to handle the fact that the precise indices of 6-frame translations depend on the length of the original nucleotide sequence, which may not be a multiple of three (as assumed in that PR).

**Issues**:
* Relates to https://github.com/ibbis-bio/common-mechanism/issues/28

### Bug fixes
* Conversion from hmmer coordinates now takes into account the translation
* No longer dependent on `transeq` from emboss, which means that biocontainer should have no warnings
* `get_query` is able to handle inexact matches in query names (e.g. when the frame is added as a suffix)

### New features
* The `Query` class now translates itself in six frames
* Set recommendation state to error at beginning of `check_biorisk`

### Refactoring
* Added a `set_recommendation` convenience method
* Did some renaming (update_from_database --> parse_hits)


alexanian:
Closing as superceded by https://github.com/ibbis-bio/common-mechanism/pull/58

PR #35, opened 2025-02-11
Urgent CLI override defaults Bugfix

MichaelBarnett:
Settings provided in CLI , with hyphens, or when provided in shorthand format (-n, instead of -skip-nt) for example, were not correctly being updated, and would not change the runs configuration from the defaults.

This fixes this issue by updating the logic on matching argv to the argsparse args object.
This was affecting the following CLI commands:
skip nt, fast mode, protein search tool, diamond jobs, all shorthand variations.

## Changes
Side effect includes the storage of shorthand and longhand command line arguments to be stored in a constant dictionary within screenio.py. This was done to ensure that the declaration of args within screen.py wasn't hard coding its own definitions, and connects the declaration directly to the logic used to detect those args - this was deemed preferable to passing the parser object, instead of args, down the call chain from run() to screenio.py init.

### Bug fixes
many CLI commands now correctly override defaults.

### Refactoring
Shorthand and longhand CLI commands for screen now stored in screenio.py dict.

### DIFF
--- a/commec/cli.py
+++ b/commec/cli.py
@@ -13,7 +13,6 @@ Command-line usage:
     - commec flag /path/to/directory/with/output.screen 
     - commec split input.fasta
 """
-import argparse
 from commec.flag import (
     DESCRIPTION as flag_DESCRIPTION,
     add_args as flag_add_args,
@@ -23,6 +22,7 @@ from commec.screen import (
     DESCRIPTION as screen_DESCRIPTION,
     add_args as screen_add_args,
     run as screen_run,
+    ScreenArgumentParser
 )
 from commec.split import (
     DESCRIPTION as split_DESCRIPTION,
@@ -35,12 +35,11 @@ from commec.setup import (
     run as setup_run,
 )
 
-
 def main():
     """
     Parse the command line arguments and call the relevant sub-command.
     """
-    parser = argparse.ArgumentParser(
+    parser = ScreenArgumentParser(
         prog="commec", description="Command-line entrypoint for the Common Mechanism"
     )
     subparsers = parser.add_subparsers(dest="command")
@@ -62,6 +61,7 @@ def main():
     setup_add_args(setup_parser)
 
     args = parser.parse_args()
+
     if args.command == "screen":
         screen_run(args)
     elif args.command == "flag":
--- a/commec/config/io_parameters.py
+++ b/commec/config/io_parameters.py
@@ -20,6 +20,20 @@ from yaml.parser import ParserError
 from commec.config.query import Query
 from commec.config.constants import DEFAULT_CONFIG_YAML_PATH
 
+SCREEN_ARGS = {
+    "database_dir"       : ("-d", "--databases"),
+    "config_yaml"        : ("-y", "--config"),
+    "fast_mode"          : ("-f", "--fast"),
+    "protein_search_tool": ("-p", "--protein-search-tool"),
+    "skip_nt_search"     : ("-n", "--skip-nt"),
+    "threads"            : ("-t", "--threads"),
+    "diamond_jobs"       : ("-j", "--diamond-jobs"),
+    "output_prefix"      : ("-o", "--output"),
+    "cleanup"            : ("-c", "--cleanup"),
+    "force"              : ("-F", "--force"),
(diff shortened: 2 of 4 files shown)


alexanian (changes requested):
Thank you for catching this! I feel so silly for failing to notice this error.

That said: I think we can make our implementation of this a bit more straightforward by using more `ArgumentParser` features directly. In particular, I think we can leave the `add_argument` logic alone (which I am inclined to do; I feel like someone reading the code where the arguments are added should be able to write a full CLI call) if we do something like this:

```
class ScreenArgumentParser(argparse.ArgumentParser):
    """Argument parser that returns a `user_specified_args` namespace item, which helps selectively override other configuration (e.g. provided via YAML)"""
    
    def parse_args(self, args=None, namespace=None):       
        # Get argument strings; in most cases, args and sys.argv[1:] will be the same
        cli_strings = args if args is not None else sys.argv[1:]
        
        user_specified_args = set()
        # self._actions has every argument that can be provided to the parser
        for possible_arg in self._actions:
            for arg_string in possible_arg.option_strings:
                if arg_string in cli_strings:
                    # Store the destination name
                    user_specified_args.add(possible_arg.dest)
        
        ns = super().parse_args(args, namespace)
        ns.user_specified_args = user_specified_args
        return ns
```

This would then enable us to do something like:

```        
    # Load user config YAML if provided
    if args.config_yaml:
        self._lazy_update_from_yaml(args.config_yaml)
    
    # Update config only with explicitly provided CLI arguments
    for arg_name in args.user_specified_args:
        value = getattr(args, arg_name)
        # Map certain argument names to different config keys
        config_key = {
            'fast_mode': 'in_fast_mode',
            'cleanup': 'do_cleanup'
        }.get(arg_name, arg_name)
``` 

What do you think?

MichaelBarnett:
taking advantage of inheritance and using it like a cheeky little wrapper, great idea. That is a much cleaner solution - I'll implement!

alexanian:
I also realized whilst I was looking at this that we have several incorrect type annotations, where `args: argparse.ArgumentParser` is actually `args: argparse.Namespace` (which is what's returned by `parse_args`). We could also avoid passing around the `Namespace` and instead be giving a `vars(args)` to `run`, `ScreenIO`, etc... but my inclination is that's it's better to leave it as the `argparse.Namespace`

MichaelBarnett:
Sorry, got to this eventually. Subparsers and the cli.py entry point made this slightly more complicated than the above fix, but ultimately its working now.

I've put the screenparser declaration in screen.py after moving it around a bunch. It can't go in clip.py else we instil circular dependencies with cli and screen.

It is used for flag / split / setup etc. Due to how everything is subparsers of each other.

Tempted to rename it to CommecParser, but ultimately it exists for screen so I'm torn haha.

Enjoy, I know this bugfix is a blocker for Brittainy so hopefully we can get this through quick.

alexanian:
Just to make sure I fully understand the hierarchy of how parameters are set:

1. (highest priority) CLI arguments specified by directly the user
2. Contents of a YAML file provided through the CLI (i.e. as an argument to `--config`
3. Contents of the default YAML file included in the package (at `DEFAULT_CONFIG_YAML_PATH`)
4. (loweset priority) CLI argument defaults that are not included in the default YAML

Am I right in thinking that we have nothing in category 4, and should perhaps remove defaults from the CLI?

MichaelBarnett:
Yes, I don't think category 4 exists in a functional sense, I believe many of the defaults have been stripped from their declarations already?

But otherwise yes, CLI > Provided Yaml > Default Yaml is the hierarchy of config input.

There is an operational defaults exception, for db_dir, aka -d --database, which requires a none default, due to how it is used to override the base path within the yaml dict. When it is passed to _format_config_paths() within io_parameters. Due to how this config can affect every path in the yaml.

Also the CLI input yaml, defaults to "", designed to fail a is valid file check, which will allows overwriting the default yamls. Other than that, all defaults for the other CLI already do not exist - i.e. no category 4.

There is also the exception of output prefix, which isn't a yaml option at this time.

alexanian (approved):
I think we're good to merge, would appreciate a double-check from you on the tests I wrote @MichaelBarnett

PR #36, opened 2025-02-24
Simple bug-fix for --protein-search-tool to the double hypen.

MichaelBarnett:
## Background
-protein-search-tool should be --protein-search-tool.

**Issues**:
Current issue with expected double hyphen usage not applying to this single CLI command.

## Changes
Added double hyphen to --protein-search-tool


ibbis-org-admin (approved):
Good catch!

PR #41, opened 2025-03-03
Diamond Job Fix.

MichaelBarnett:
## Background
Calculating the optimal number of concurrent diamond jobs, and threads per job was being handled when no jobs were provided, however was incorrectly working when the user forced a non-ideal job number as input. This addresses better controls of user job input, allowing for non-optimal job count testing, as well as ensuring commec keeps to the thread limit.

## Changes
The number of input diamond jobs (i.e. concurrent diamond runs) is interrogated to be both less than the number of diamond database files, and to be less than the number of available threads. This also occurs before the calculation to determine number of threads per job.

User input of desired job count, will override better greatest common denominator solutions, at the cost of potentially not utilising all available (max threads) assigned threads.

### Bug fixes
The number of threads calculated for each job will no longer be based off the a job number which is simply too large for the program to handle. i.e. when jobs > max_threads, or jobs > number of databases.

Unit testing for checking these calculations are updated to include None, and 1-10 input user jobs.


commec/tests/test_dbs.py
```diff
@@ -92,9 +96,22 @@ def test_diamond_job_and_threads_calculations():
...
+                  n_database_files, "dbs, and", 
+                  max_threads, "max threads: Do", 
+                  threads_per_run, "threads per", concurrent_runs,"runs.")
+
+            # If input jobs is provided, we should never exceed max threads.
+            assert concurrent_runs * threads_per_run <= max_threads
+
+            # If no number of input jobs is provided, we should ALWAYS use all available threads.
+            if input_jobs is None:
+                assert concurrent_runs * threads_per_run == max_threads
```
alexanian, line 117: Could we rewrite this test to be more explicit about expected results? That will help us pinpoint errors if we mess something up in future, and will also let us cut out the debug printing. Maybe something like:

```
@pytest.mark.parametrize(
    "input_jobs, max_threads, n_database_files, expected_runs, expected_threads",
    [
        (None, 20, 6, 2, 10), # jobs capped by db count, using all threads
        (None, 8, 5, 1, 5),   # jobs capped by db count, not using all threads
        (3, 12, 6, 3, 4),     # jobs=3 --> 3 runs with 4 threads each
        (10, 20, 5, 5, 4),    # jobs=10 > db=5, capped to 5 runs with 4 threads each      
        (20, 10, 5, 5, 2),    # jobs=20 > threads=10, capped to 5 runs with 2 threads each
        (10, 4, 5, 6, 1),     # jobs=10 > db, threads, cappted to 4 runs with 1 thread each  
    ]
)
def test_diamond_job_and_threads_calculations_parametrized(
    input_jobs, max_threads, n_database_files, expected_runs, expected_threads
):
```

I think we might also want to test the `jobs = 0` case, I'm not 100% sure we handle that properly right now.

alexanian (changes requested):
I think we'd discussed also setting

    n_concurrent_runs =  min(number_of_databases, max_threads)

instead of using GCD, based on Brittany's test results, but I'm happy for us to hold that off for the larger threading updates?

commec/tests/test_dbs.py
MichaelBarnett: Fair enough, after incorporating this, and some further testing, the original test came in handy as I fiddled with the business logic handling some niche cases. Thus I kept the other test too.

MichaelBarnett:
Just noting I agree and think its a good idea to put all threading number decisions (such as changing the use of GCD) in a single feature branch that incorporates Brittany's data for all database tools, not just Diamond.

commec/tests/test_dbs.py
```diff
@@ -90,6 +90,9 @@ def test_diamond_job_and_threads_calculations(input_jobs):
     Tests a range of threads, and diamond database sizes,
     for automatically calculating the optimum number of runs,
     and threads per run. Such that no CPU time is wasted.
+    No specific expected outcomes, but we can check general expectations:
+     - Never exceed max threads.
+     -
```
alexanian, line 95:
```suggestion
     No specific expected outcomes, but we check general expectations (e.g. never exceeding max_threads)
```

alexanian (approved):

PR #30, opened 2025-01-30
Debugging v 0.3

MichaelBarnett:
## Background

Fixes various minor bugs to be ready for the 0.3.0 release, important for posterity sake before json overhaul.

## Changes
* Write up for adding debugging into main (pre json merge)
* Minimum python version consistently set to 3.1
* Removed typing features to be deprecated after python 3.9
* Updated config yaml as packaged resource - should work properly for conda recipes and environments.
* Improved human readable outputs on cli inputs.
* Improved Diamond warnings.
* Removed unused packages.
* Updated how default configuration is handled through default yamls.
* Outputs the yaml configuration used per run.
* Improved output folder hierarchy.

### New features
* Outputs the yaml config to allow reproducibility of settings.

### Breaking changes 
* Alters the output file structure from the expected single folder dump - may break analysis dev tools which don't expect the new file structure.

PR #42, opened 2025-03-04
Restore non-directory-based output format for v0.3

alexanian:
## Background

We have big plans to update the output format for commec, so that outputs are organized into folders with the following hierarchy:

* {queryname}
  * {queryname}.screen.log
  * {queryname}.json
  * {queryname}.html
   * {queryname}_input
       * _dir of generated input files, e.g. {queryname}.cleaned.fasta_
   * {queryname}_output
       * _dir of intermediate output files, e.g. {queryname}.nt.blastn_

We did a version of this in https://github.com/ibbis-bio/common-mechanism/pull/30/, but realized that this was too breaking of a change for a minor update (e.g. `commec flag` made assumptions about the output structure, we were a little too aggressive in our approach to shortening filenames).

As a result, this undoes that change (which we will likely reintroduce in v1.0) and adds a dev-script, `collate-screens.py`, which allows reformatting outputs from v0.3 to a flags-ingestible format.

### Bug fixes
* Expand variables and avoid creating directories named "~"

### New features
* Restore previous output format, including handling of directory inputs
* Add collate-screens dev script


MichaelBarnett (approved):
Appears to work as intended for me.

PR #50, opened 2025-04-08
Json rebase

MichaelBarnett:
## Background
The JSON and HTML outputs branch has had an extensive history of development, which has been condensed and rebased onto the "Develop" branch as it is too extensive to squash merge as a single PR.

At this stage, the JSON output branch represents Commec version 1.0, and has the following list of changes (above the existing changes present in 0.3.2):

## Changes
* Commec now outputs a `.json`, and `.html`, instead of a `.screen` file.
* `.screen`-like output can still be found within `.log`
* Overall outputs are re-arranged into a `/name_output/` `/name_input/` `name.json` `name.log` `name.html` file structure. Automatically generated inputs are within the input folder, and raw outputs from each step are in the outputs folder.
* Commec output is now query centric, with additional support for multi-fasta files, across all steps of screening.

### Bug fixes
* Indexing Corrections for HMMER searches, where the AA index is now correctly converted from 6 frame translations to query nucleotide coordinates.
* Indexing Corrections for non-coding regions that are parsed specifically to Nucleotide Taxonomy, are correctly converted back to query nucleotide coordinates.
* Benign clearing logic now correctly applies to the region affecting the hit, rather than the length of the entire query. ( > 80% coverage etc)
* Fixed an issue where externally threading Commec could cause a race condition in file creation.

### New features
* JSON output, which is always at least partially generated on failure, contains a database summary of all pertinent aspects of a Commec Screen Run, for easy external parsing. The expected structure is found in `Commec/config/result.py`. Includes overall status from each step (Similar to Commec flag).
* A visual summary is present in the output HTML, which shows all queries as a visual graph, aligned against each hit, colour coded according to hit type, and status.
* Significant improvements to logging and additional verbosity in debug logging.
* Improved identifying and collapsing of similar-hits to avoid hit overload in some taxonomy search scenarios.
* Much greater unit test coverage, including a full screen functional test on mock data. Expected outputs can be updated by passing --gen-examples to pytest.
### Breaking changes 
* The output folder structure has changed, which breaks scripts expecting the old all-in-one folder organisation.

### Refactoring
* The output JSON database structures represent in memory the current state of a screen run, on which the screening logic can be applied at any point of the screen process.
* All Check_X functions have been replaced with JSON state alternatives.


alexanian (approved):
Since this is a +4166 -629 LOC PR, I've already reviewed various parts of it over its development, no major errors popped up in my functional tests, *and* we're going to do additional functional tests + features before we actually release this from develop, I think we should merge this now so that we can start new features (for v1.0, let's go!) from it.

PR #52, opened 2025-04-18
Output formatting fixes

MichaelBarnett:
## Background
After the great rebase of JSON and HTML functionality into develop, this PR addresses a first round of feedback with several fixes.

## Changes
* fixed typos of Virulence.
* Updated non-verbose logging to be more consistent across steps, and at a per query level.
* All databases/search/screen are consistently referenced as Search, both in outputs and internally.
** (The only Screen reference is to the Commec Screen itself, a Commec Screen can be thought of being composed of many Searches)
* Additional regulation annotations now referred to as regulated_taxonomy in taxonomy searches.
* JSON output now includes a small query_info section, containing an abs path to the original fasta file. Query number, and total length in bps of all queries.
* JSON query no longer stores the full sequence information.
* HTML page now has custom heading functionality for the tab.
* Beautified the HTML outputs.
* Taxonomy hits which share exactly the same region coordinates on the query now share a hit, and update their information. Leading to better collapsing of overly verbose outputs.

### Bug fixes
* Important bug fix to query name key generation to avoid issues with translation truncated names with the | character.
* Skip is no longer used as a potential benign state to override overall recommendation.
* Fixed an issue where the incorrect query length was being used to calculate the final non-coding region. Updated functional test to no longer unexpectedly fail for nucleotide due to this.

* Debug --verbose mode now prints all the names, and original names for all queries.

### DIFF
--- a/commec/config/query.py
+++ b/commec/config/query.py
@@ -51,13 +51,20 @@ class Query:
             )
 
     @staticmethod
-    def create_id(name : str) -> str:
+    def create_id(input_name : str) -> str:
         """
         Parse the Fasta SeqRecord string ID into 
         a 25 digit maximum Unique Identification.
         For internal Commec Screen Use only.
         Original Fasta name IDs are used during JSON output.
         """
+        # Protections on split name conventions for translation output:
+        name = input_name.split('|')[0]
+
+        # Protection on accidentally ending with _
+        if name.endswith("_"):
+            name = name[:-1]
+
         if len(name) < 26:
             return name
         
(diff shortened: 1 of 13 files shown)


MichaelBarnett:
Note that this PR is not addressing some of the finer points of storing the additional "mixed" taxonomy information as discussed. Which will require further thought, likewise the hover-text has not been updated in html until those changes are finalised, as I suspect all of that will require more finicky back and forth.

alexanian (approved):
We might make more changes to the output formats, but this package of changes all LGTM!

PR #54, opened 2025-05-01
Add --version, -v to commec cli.

MichaelBarnett:
Simple adds in the ability to print the current Commec Version being used.

Example output terminal text:

```
Commec  : The Common Mechanism
Version : 0.3.2
Copyright IBBIS (c) 2021-2025
International Biosecurity and Biosafety Initiative for Science
```

### DIFF
--- a/commec/__init__.py
+++ b/commec/__init__.py
@@ -1,0 +1,5 @@
+from importlib.metadata import version
+try:
+    __version__ = version(__name__)
+except:
+    __version__ = "X.X.X"
--- a/commec/__init__.py
+++ b/commec/__init__.py
@@ -1,5 +1,5 @@
-from importlib.metadata import version
+from importlib.metadata import version, PackageNotFoundError
 try:
-    __version__ = version(__name__)
-except:
+    __version__ = version("commec")
+except ImportError, PackageNotFoundError:
     __version__ = "X.X.X"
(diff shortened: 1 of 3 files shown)


alexanian (commented):
Should we be setting the `__version__` attribute? Looking at https://packaging.python.org/en/latest/discussions/versioning/#accessing-version-information-at-runtime

> Many projects also choose to version their top level [import packages](https://packaging.python.org/en/latest/glossary/#term-Import-Package) by providing a package level `__version__` attribute:
```
import cryptography
cryptography.__version__
'41.0.7'
```

I also wonder if we should locate that somewhere other than `config.result`, since in general config is only used by `commec screen`, but the version is useful for any commec commands?

MichaelBarnett:
Updated so we get the __version__ in __init__, which is a much better place to store and get that information I agree.

commec/__init__.py
```diff
@@ -0,0 +1,5 @@
+from importlib.metadata import version
+try:
+    __version__ = version(__name__)
+except:
+    __version__ = "X.X.X"
```
alexanian, line 5: I think directly referencing the package name is preferable (just in case this is imported in a weird way_

```suggestion
from importlib.metadata import version, PackageNotFoundError
try:
    __version__ = version("commec")
except ImportError, PackageNotFoundError:
    __version__ = "X.X.X"

```

alexanian (approved):

PR #55, opened 2025-05-01
Reject querys below minimum length

MichaelBarnett:
## Background
Short sequences are a thorn in screenings sensitivity side. It is best to ensure we have a hard line in the sand on a minimum viable sequence size below which Commec Screen should simply not be used.

## Changes
Any sequences which are less than the constant, currently set to 40 nucleotides, are rejected at the import fasta phase. A warning is given to the user about any rejected sequences. Commec will still try to run if other sequences were valid.

## Refactors
To note, the function `is_valid_record` of query validation is statically part of ScreenIO, and could be used to include any future validation we want to have on Query SeqRecord objects.

### DIFF
--- a/commec/config/constants.py
+++ b/commec/config/constants.py
@@ -2,3 +2,6 @@
 # Copyright (c) 2021-2024 International Biosecurity and Biosafety Initiative for Science
 
 DEFAULT_CONFIG_YAML_PATH = "screen-default-config.yaml"
+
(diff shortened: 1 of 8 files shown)


alexanian:
Based on our most recent conversation, plan is to merge https://github.com/ibbis-bio/common-mechanism/pull/57 and build off that, because updating this logic got messy, so closing this for now.

PR #56, opened 2025-05-01
Formatless logging

MichaelBarnett:
## Background
Sometimes, you need to copy paste a url, or console command from debug mode, and the beautiful formatting for the logs was making that issuesome. Furthermore, the final table summary print at the end of a screen run could benefit from not being beholden to the formatting system used in logging. Thus we add a way for the formatting to be removed, and altered, from the system currently in place.

## Changes
Any logger function can now parse some keywords within the extra dictionary, which allows custom logging behaviour - mainly to remove suffix log formatting , and optionally neatly cap off the existing formatting.

## Features
All console outputs which feature a some copy/paste convenient output, are logged on a single line, with no maximum width, to make copy/pasting easy for the user.


commec/utils/logger.py
```diff
@@ -44,7 +43,32 @@ def _find_message_start(self):
         return sample_formatted.find(sample.msg)
 
     def format(self, record):
+        """
+        Custom formattor for Commec purposes.
+        Accepts keywords in the extra dictionary:
+        one_line : skips text wrapping.
+        no_fmt : Skips the "WARNING | " prefixes.
+        cap, cap_up, cap_down : Neatly ties off formatted prefixes.
```
alexanian, line 51:
```suggestion
        Custom formatter for Commec logging.  Accepts the following keywords in the extra dictionary:
        no_wrap : skips text wrapping.
        no_prefix: Skips the "%(asctime)s│ %(levelname)-8s│ % " prefixes.
        box, box_up, box_down: Use unicode box-drawing characters ( "─┘" or "─┐") to tie off formatted prefixes when switching to no-prefix lines.
```
(this makes it clearer to me what's going on, but if you feel like this names are less clear, feel free to reject)

alexanian (changes requested):
small request that's just about wording

PR #57, opened 2025-05-02
Rationale

MichaelBarnett:
## Background
Whilst looking at a list of FLAG, WARN, and PASS screen status' is nice, what is missing is the rationale behind these decisions in a human readable format. This adds an initial implementation of rationale to the recommendation container of a query, and gives a quick grab text, from `screendata.query.recommendation.rationale`, for providing the user or tester to read.

## Changes
* A rational text is determined during screen data update.
* Screen data update logic has been improved to be more robust.
* Benign steps are no longer always PASS, and reflect the status of what they did. e.g. Flag (Cleared)
* The summary table is now printed with only the final screen status, and the rational, instead of each step.
* The HTML has been updated to include the rationale in the headings, and in the hover over text of the query bar.
* Each query has a new boolean flag, `no_hits_warning`, initialised to `True`, as commec screen runs, the query is updated to check for whether it has hit _anything_ at all. When updating screen data, if a query is still True, it will marked as a warning (instead of PASS from all steps) and the rationale being there is no known homology. 
** Note: a future implementation can build on this behaviour and detect poorly represented regions rather than treat the query as a whole.
** Note: This warning also works well to warn against short sequences that are too short to find any meaningful hits, to this end, we can alter the rationale to include (query likely too short) when the query length is small, rather than skipping those queries altogether as per #55. Is that PR therefore redundant and this a better solution?
 
### Bug fixes
* Fixed an off-by-one graphic glitch in HTML bar rendering.

### New features
* Rationale is added in Commec Outputs.


commec/config/result.py
```diff
@@ -504,18 +569,19 @@ def get_flag_data(self) -> pd.DataFrame:
...
                 "overall": query.recommendation.screen_status,
-                "biorisk": query.recommendation.biorisk_status,
-                "taxonomy_aa": query.recommendation.protein_taxonomy_status,
-                "taxonomy_nt": query.recommendation.nucleotide_taxonomy_status,
-                "benign": query.recommendation.benign_status
+                "rationale": query.recommendation.rationale,
+                #"biorisk": query.recommendation.biorisk_status,
+                #"taxonomy_aa": query.recommendation.protein_taxonomy_status,
+                #"taxonomy_nt": query.recommendation.nucleotide_taxonomy_status,
+                #"benign": query.recommendation.benign_status
```
alexanian, line 576: Is this intentional?

commec/config/result.py
```diff
@@ -260,28 +261,52 @@ def update_query_flag(self):
             self.benign_status
         }:
             self.screen_status = ScreenStatus.ERROR
+            self.rationale = "There was an error during screening"
```
alexanian, line 264: Is it possible for us to say which step had the error here?

commec/config/result.py
```diff
@@ -260,28 +261,52 @@ def update_query_flag(self):
...
+            return
+                
+        if self.benign_status == ScreenStatus.CLEARED_FLAG:
+            self.screen_status = self.benign_status
+            self.rationale = "A sequence of concern was cleared as benign"
```
alexanian, line 269: Since we've learned that "benign" is kind of an obscure word for some non-native English speakers, what about:

```suggestion
            self.rationale = "No regions of concern; all earlier flags were cleared as common or non-hazardous"
```

commec/config/result.py
```diff
@@ -260,28 +261,52 @@ def update_query_flag(self):
...
+            return
+
+        if self.nucleotide_taxonomy_status == ScreenStatus.FLAG:
             self.screen_status = ScreenStatus.FLAG
+            self.rationale = "A regulated nucleotide sequence was detected"
             return
 
         if self.biorisk_status == ScreenStatus.WARN:
             self.screen_status = ScreenStatus.WARN
+            self.rationale = "A known virulance factor was detected"
```
alexanian, line 299: I think this actually makes it sound more concerning than it is. Maybe:
```suggestion
            self.rationale = "Matched to a virulence factor, but these are often shared between regulated and non-regulated organisms"
```

commec/config/result.py
```diff
@@ -260,28 +261,52 @@ def update_query_flag(self):
...
+            return
+
+        if self.benign_status == ScreenStatus.PASS:
             self.screen_status = self.benign_status
+            self.rationale = "No regulated regions of concern"
             return
 
         if self.biorisk_status == ScreenStatus.FLAG:
             self.screen_status = ScreenStatus.FLAG
+            self.rationale = "A known biorisk was detected"
```
alexanian, line 284: I know we eventually want to propagate up more detail (toxin? viral virulence factor?) but for now maybe
```suggestion
            self.rationale = "Matched a sequence with pathogenic or toxic function "
```

commec/config/result.py
```diff
@@ -260,28 +261,52 @@ def update_query_flag(self):
...
+            self.rationale = "A sequence of concern was cleared as benign"
             return
 
-        if self.benign_status not in {
-            ScreenStatus.NULL, 
-            ScreenStatus.SKIP
-        }:
+        if self.benign_status == ScreenStatus.CLEARED_WARN:
+            self.screen_status = self.benign_status
+            self.rationale = "A mixed sequence of concern was cleared as benign"
```
alexanian, line 274: Could probably use the same benign message as above?

commec/config/result.py
```diff
@@ -260,28 +261,52 @@ def update_query_flag(self):
...
+            return
+
+        if self.benign_status == ScreenStatus.PASS:
```
alexanian, line 277: When is benign_status pass? When the overall status was already pass? I wonder if we should skip benign then...

commec/config/result.py
```diff
@@ -382,6 +410,25 @@ def _update_step_flags(self):
...
         Then updates the Query flag to consolidate all.
         """
+        logger.debug("Updating step status flags for query %s", self.query)
+        logger.debug("Current status %s", {
+            "Biorisk":str(self.recommendation.biorisk_status),
+            "Protein":str(self.recommendation.protein_taxonomy_status),
+            "Nucleot":str(self.recommendation.nucleotide_taxonomy_status),
+            "benign ":str(self.recommendation.benign_status)})
+        
+        ignored_status = {ScreenStatus.SKIP, ScreenStatus.ERROR, ScreenStatus.PASS}
```
alexanian, line 420: This section feels like it might be more readable if we added in something like the `set_recommendation` method I wrote in https://github.com/ibbis-bio/common-mechanism/pull/31/files. This is that method (but we'd want to include your `ignored_status` logic, too, I think).

```
    def set_recommendation(self, step : ScreenStep, recommendation: Recommendation):
        """
        Set the query recommendation for a specific screen step.
        """
        # Map ScreenStep to the corresponding attribute in QueryRecommendationContainer
        step_to_field = {
            ScreenStep.BIORISK: 'biorisk_screen',
            ScreenStep.TAXONOMY_NT: 'nucleotide_taxonomy_screen',
            ScreenStep.TAXONOMY_AA: 'protein_taxonomy_screen',
            ScreenStep.BENIGN_PROTEIN: 'benign_screen',
            ScreenStep.BENIGN_RNA: 'benign_screen',
            ScreenStep.BENIGN_SYNBIO: 'benign_screen'
        }
        field_name = step_to_field.get(step)
        current = getattr(self.recommendation, field_name)
        new_rec = compare(current, recommendation)
        setattr(self.recommendation, field_name, new_rec)
 ```

**Review by alexanian — CHANGES_REQUESTED** [review]
Added a ton of small proposed wording changes, angling towards giving decision guidance vs. factual info.

I still think it's a good idea to skip short queries ― I think we should look for opportunities to skip processing effort wherever possible. I do like the idea of, in future, showing which regions have "no hits" ― maybe as part of the taxonomy overhaul, where I think it would be great to include a taxonomic best guess for all regions, not just flagged ones?

File: commec/config/result.py [hunk 9]
```diff
@@ -297,6 +322,9 @@ def update_query_flag(self):
         if self.screen_status == ScreenStatus.NULL:
             self.screen_status = self.biorisk_status
 
+        if self.screen_status == ScreenStatus.PASS:
+            self.rationale = "No regulated regions of concern"
```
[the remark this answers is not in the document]
MichaelBarnett: I agree with this idea - of switching to a standard of risk tiers - however that would definitely be its own PR.

commec/config/result.py
MichaelBarnett: Yes, for the moment this is intentional, wanted to communicate how this summary table would be changed in the sense of missing info. We need to ask the question - where do people see this (cli/logs only, in automated pipelines probably not at all), and what info do we actually want here.

What are your thoughts? I am leaning towards almost printing two tables, the 1st one showing a "flag"-like output, the original.
And the 2nd only showing the truncated reasoning?

commec/config/result.py
MichaelBarnett: Why not! However it will only be the "first" error for now.

commec/config/result.py
```diff
@@ -382,6 +410,25 @@ def _update_step_flags(self):
         Updates the steps within QueryScreenStatus to be congruent for every hit.
         Then updates the Query flag to consolidate all.
         """
+        logger.debug("Updating step status flags for query %s", self.query)
+        logger.debug("Current status %s", {
+            "Biorisk":str(self.recommendation.biorisk_status),
+            "Protein":str(self.recommendation.protein_taxonomy_status),
+            "Nucleot":str(self.recommendation.nucleotide_taxonomy_status),
```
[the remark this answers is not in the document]
MichaelBarnett: Yeap! Prettified.

commec/config/result.py
MichaelBarnett: Absolutely

commec/config/result.py
MichaelBarnett: hangover from previous behaviour, removed.

MichaelBarnett:
Updates to address comments: New example output:
```
 >> EXPORT JSON SUMMARY :
                    query      overall      biorisk    taxonomy_aa    taxonomy_nt       benign
     SUPER_SHORT_QUERY_40         Skip         Skip           Skip           Skip         Skip
           SHORT_QUERY_50      Warning         Pass           Pass           Pass         Pass
BBa_I766605_YopH-EE_under      Warning      Warning Flag (Cleared)           Pass      Warning
         YopH-EE-Reverse1      Warning      Warning Flag (Cleared)           Pass      Warning
         YopH-EE-Reverse2      Warning      Warning Flag (Cleared)           Pass      Warning
         YopH-EE-Reverse3      Warning      Warning Flag (Cleared)           Pass      Warning
 AY578694.1_African_swine         Flag         Pass           Flag           Pass         Flag
      BBa_K133012_A_13220         Flag         Flag Flag (Cleared) Flag (Cleared)         Flag
   red-team_500bp_Ricinus         Flag         Flag           Pass           Skip         Flag
 BBa_K4781000_Xanthomonas      Warning      Warning           Pass           Pass      Warning

 >> RATIONALE :
SUPER_SHORT_QUERY_40      : Skip         --> Query is too short, and was skipped.
SHORT_QUERY_50            : Warning      --> No matches to sequences from any step. Unknown or potential risk factor. Sequence may be too short, or in silco de novo.
BBa_I766605_YopH-EE_under : Warning      --> Matches sequence with virulence factor; as well as flags cleared as common or non-hazardous.
YopH-EE-Reverse1          : Warning      --> Matches sequence with virulence factor; as well as flags cleared as common or non-hazardous.
YopH-EE-Reverse2          : Warning      --> Matches sequence with virulence factor; as well as flags cleared as common or non-hazardous.
YopH-EE-Reverse3          : Warning      --> Matches sequence with virulence factor; as well as flags cleared as common or non-hazardous.
AY578694.1_African_swine  : Flag         --> Matches protein sequence with regulated organisms.
BBa_K133012_A_13220       : Flag         --> Matches sequence with pathogenic or toxin function; as well as virulence factor; as well as flags cleared as common or non-hazardous.
red-team_500bp_Ricinus    : Flag         --> Matches sequence with pathogenic or toxin function.
BBa_K4781000_Xanthomonas  : Warning      --> Matches sequence with virulence factor.

```

commec/config/result.py
```diff
@@ -245,12 +288,14 @@ class QueryScreenStatus:
     protein_taxonomy_status: ScreenStatus = ScreenStatus.NULL
     nucleotide_taxonomy_status: ScreenStatus = ScreenStatus.NULL
     benign_status: ScreenStatus = ScreenStatus.NULL
+    rationale : str = "-"
 
-    def update_query_flag(self):
+    def update_screen_status(self, query_data):
```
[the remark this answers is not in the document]
MichaelBarnett: Just remembered reading through your use of max() earlier part of the reason why I was over-writing operators: This works extremely well:

````python
    def update(self, query_data):
        """
        Parses the query status across all screening steps, 
        then updates overall status flag.
        """

        self.screen_status = max(
            self.biorisk,
            self.protein_taxonomy,
            self.nucleotide_taxonomy,
            self.benign
        )

        # If everything is happy, but we haven't hit anything, time to be suspicious...
        if (self.screen_status == ScreenStatus.PASS and query_data.no_hits_warning):
            self.screen_status = ScreenStatus.WARN
            return

        # If biorisk is skipped then it is skipped overall - likely query is too short...
        if (self.biorisk == ScreenStatus.SKIP):
            self.screen_status = ScreenStatus.SKIP
            return
```

commec/config/result.py
```diff
@@ -259,41 +304,89 @@ def update_query_flag(self):
...
+
+        # If biorisk is skipped then it is skipped overall - likely query is too short...
+        if (self.biorisk_status == ScreenStatus.SKIP):
+            self.screen_status = ScreenStatus.SKIP
+            return
+
+        logger.debug("Update called on recommendation, however no issuesome status updates occured."
+                     " Setting status to Biorisk output [%s] (Should be PASS, or NULL)",
+                     self.biorisk_status)
+        assert self.biorisk_status in [ScreenStatus.PASS, ScreenStatus.NULL]
```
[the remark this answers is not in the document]
MichaelBarnett: Definitely all addressed in latest PR merge and commits! =)

commec/config/result.py
```diff
@@ -324,75 +324,31 @@ def _get_step_field_and_status(self, step: ScreenStep) -> tuple[str, ScreenStatu
...
-            return
-
-        if self.protein_taxonomy == ScreenStatus.WARN:
-            self.screen_status = ScreenStatus.WARN
-            return
-
-        if self.nucleotide_taxonomy == ScreenStatus.WARN:
-            self.screen_status = ScreenStatus.WARN
-            return
+        )
```
alexanian, line 340: lolll so efficient hahaha

alexanian (approved):

PR #58, opened 2025-05-19
Emboss removal

MichaelBarnett:
## Background
Recreation of #31, which is a little old compared to the latest develop features which also had significant crossover with all #31 features and changes, and also incorporates some additional refactor complications.

The  `biocontainer` auto-built from the `commec` bioconda packages [currently has some gnarly security warnings](https://quay.io/repository/biocontainers/commec) (related to https://github.com/bioconda/bioconda-recipes/issues/46980#issuecomment-2604135864 ). This removes that dependency by writing our own `translate` method.

Also updates #26 (I have marked this as a draft, since that one should be merged first) to handle the fact that the precise indices of 6-frame translations depend on the length of the original nucleotide sequence, which may not be a multiple of three (as assumed in that PR).

**Issues**:
* Relates to https://github.com/ibbis-bio/common-mechanism/issues/28

### Bug fixes
* No longer dependent on `transeq` from emboss, which means that biocontainer should have no warnings

### New features
* The `Query` class now translates itself in six frames

### DIFF
--- a/commec/config/query.py
+++ b/commec/config/query.py
@@ -3,6 +3,7 @@
 
 import subprocess
 from Bio.SeqRecord import SeqRecord
+from dataclasses import dataclass
 from commec.config.result import QueryResult
 
 class Query:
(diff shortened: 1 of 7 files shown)


alexanian (approved):
Approved; this is an updated version of the emboss-related code in https://github.com/ibbis-bio/common-mechanism/pull/31/, and thus this is sort of a self-review, but: let's merge it!

PR #59, opened 2025-05-20
Update default nt database to core_nt

MichaelBarnett:
## Background
Sets the default NT database used to core_nt. As this is smaller, faster, whilst providing enough sensitivity for commec screening purposes.

## Changes
core_nt now the default database, which updates the default yaml, as well as how setup downloads core_nt instead of nt.

### New features
Uses core_nt by default.

### Breaking changes 
If you already have NT installed, and rely on default database setup locations with -d cli, you may need to make a custom .yaml now.


alexanian (approved):
yessss

PR #60, opened 2025-05-27
Adds additional debug output to record input CLI and parameters used in screen.

MichaelBarnett:
## Background
We had an issue with -force result in -f -o rce, this may of been more quickly identified with more verbose debug logging. This PR aims to improve future argument and parameter visibility in a debug context.

## Changes
Prints the args.NameSpace and the screen parameters dictionary overidden by yaml after settings parsing to the debug log.

Example output:
```
 The Common Mechanism : Screen
────────┐
DEBUG   │ Parsing input parameters...
DEBUG   │ Imported the following arguments:
DEBUG   │ {'command': 'screen',
        │  'config_yaml': '../test_config.yaml\r',
        │  'database_dir': '/root/commec-dbs/',
        │  'diamond_jobs': None,
        │  'do_cleanup': False,
        │  'fasta_file': '../../../example_data/single.fasta',
        │  'force': False,
        │  'in_fast_mode': False,
        │  'output_prefix': '../test_foldseek/',
        │  'protein_search_tool': None,
        │  'resume': True,
        │  'skip_nt_search': False,
        │  'threads': 4,
        │  'user_specified_args': {'config_yaml',
        │                          'database_dir',
        │                          'output',
        │                          'output_prefix',
        │                          'resume',
        │                          'threads',
        │                          'verbose'},
        │  'verbose': True,
        │  'version': False}
DEBUG   │ Running Screen with the following parameter set:
DEBUG   │ {'base_paths': {'default': '/root/commec-dbs/'},
        │  'databases': {'benign': {'cm': {'path': '/root/commec-dbs/benign_db/benign.cm'},
        │                           'fasta': {'path': '/root/commec-dbs/benign_db/benign.fasta'},
        │                           'hmm': {'path': '/root/commec-dbs/benign_db/benign.hmm'}},
        │                'biorisk_hmm': {'annotations': '/root/commec-dbs/biorisk_db/biorisk_annotations.csv',
        │                                'path': '/root/commec-dbs/biorisk_db/biorisk.hmm'},
        │                'regulated_nt': {'path': '/root/commec-dbs/nt_igem_examples/nt_igem_examples'},
        │                'regulated_protein': {'blast': {'path': '/root/commec-dbs/pdb_nr/pdbaa'},
        │                                      'diamond': {'path': '/root/commec-dbs/nr_dmnd/nr.dmnd'},
        │                                      'protein_search_tool': 'blast'},
        │                'taxonomy': {'benign_taxids': '/root/commec-dbs/benign_db/vax_taxids.txt',
        │                             'path': '/root/commec-dbs/taxonomy/',
        │                             'regulated_taxids': '/root/commec-dbs/biorisk_db/reg_taxids.txt',
        │                             'regulated_vaxids': '/root/commec-dbs/biorisk_db/reg_taxids.txt',
        │                             'taxonomy_directory': '/root/commec-dbs/taxonomy/'}},
        │  'diamond_jobs': None,
        │  'do_cleanup': False,
        │  'force': False,
        │  'in_fast_mode': False,
        │  'protein_search_tool': 'blastx',
        │  'resume': True,
        │  'skip_nt_search': False,
        │  'threads': 4,
        │  'verbose': True}
INFO    │ Validating input query and databases...
```

### DIFF
--- a/commec/config/screen_io.py
+++ b/commec/config/screen_io.py
@@ -12,6 +12,7 @@ import argparse
 import logging
 import multiprocessing
 import importlib.resources
+from pprint import pformat
 
 import yaml
 from yaml.parser import ParserError
@@ -176,6 +177,11 @@ class ScreenIO:
 
         # Update paths in configuration using appropriate string substitution
         self._format_config_paths(self.db_dir)
+
+        logger.debug("Importing the following argument namespace:")
+        logger.debug(pformat(vars(args)))
+        logger.debug("Running Screen with the following parameter set:")
+        logger.debug(pformat(self.config))
 
     def _load_config_from_yaml(self, config_filepath: str | os.PathLike) -> dict:
         """
--- a/commec/config/screen_io.py
+++ b/commec/config/screen_io.py
@@ -170,6 +170,7 @@ class ScreenIO:
         # Override configuration with any in user-provided YAML file
         cli_config_yaml=args.config_yaml.strip()
         if os.path.exists(cli_config_yaml):
+            logger.debug(f"Overriding defaults in {default_yaml} with values from {cli_config_yaml}")
             self._update_config_from_yaml(cli_config_yaml)
 
         # Override configuration with any user-provided CLI arguments
@@ -178,8 +179,6 @@ class ScreenIO:
         # Update paths in configuration using appropriate string substitution
         self._format_config_paths(self.db_dir)
 
-        logger.debug("Importing the following argument namespace:")
-        logger.debug(pformat(vars(args)))
         logger.debug("Running Screen with the following parameter set:")
         logger.debug(pformat(self.config))
 
@@ -216,8 +215,12 @@ class ScreenIO:
             )
 
         # Update the YAML default values in the configuration dictionary
+        logger.debug("Using the following CLI configuration arguments:")
+        logger.debug(pformat(args.user_specified_args))
+
         for arg in args.user_specified_args:
-             if arg in self.config and hasattr(args, arg):
+            if arg in self.config and hasattr(args, arg):
+                logger.debug(f"Command line arguments updated {arg} to {getattr(args,arg)}")
                 self.config[arg] = getattr(args, arg)
 
     def _format_config_paths(self,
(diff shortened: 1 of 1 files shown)


commec/config/screen_io.py
```diff
@@ -177,6 +178,11 @@ def _read_config(self, args: argparse.Namespace):
         # Update paths in configuration using appropriate string substitution
         self._format_config_paths(self.db_dir)
 
+        logger.debug("Importing the following argument namespace:")
+        logger.debug(pformat(vars(args)))
```
alexanian, line 182: I worry there is something potentially confusing about logging the entire argument namespace, because only user-specified arguments are actually used for the override. 

I think I would propose that we move this logging statement to the `_update_config_from_cli` method, and then replace:

```python
        # Update the YAML default values in the configuration dictionary
        for arg in args.user_specified_args:
             if arg in self.config and hasattr(args, arg):
                self.config[arg] = getattr(args, arg)
```

with
```python
    # Update the YAML default values in the configuration dictionary
     overrides = {k: v for k, v in args.items() if k in args.user_specified_args}
     logger.debug("Using the following CLI configuration arguments:")
     logger.debug(pformat(vars(overrides)))
    for arg, value in overrides.items():
        if arg in self.config:
            self.config[arg] = value
```

alexanian (changes requested):
One suggestion, but makes sense to me! While doing this, can we also add
```python
    if os.path.exists(cli_config_yaml):
            logger.debug(f"Overriding defaults in {default_yaml} with values from {cli_config_yaml}")
            self._update_config_from_yaml(cli_config_yaml)
 ```

File: commec/config/screen_io.py (hunk 1 above)
Comment (MichaelBarnett) [reply commec/config/screen_io.py:182]: So i hear you, but at the time figured it might be useful to also show the defaults, I think rather than changing the logic too much here (I don't know if args behaves like a dict with an .items() function as its the argsparse.Namespace, and that overrides = line is less than fun to read.

Perhaps we just add the log within the existing loop:
```
for arg in args.user_specified_args: 
    if arg in self.config and hasattr(args, arg):
        logger.debug(f"Updating {arg} = {getattr(args, arg)}")
        self.config[arg] = getattr(args, arg)
```

Comment (alexanian) [reply commec/config/screen_io.py]: Plausibly *I* am confused here, but if (for example) we messed up keeping the config aligned, such that `screen-default-config.yaml` had:

```
do_cleanup: True
```
while we still had
```
    output_handling_group.add_argument(
        "-c",
        "--cleanup",
        dest="do_cleanup",
        action="store_false",
        help="Delete intermediate output files for this Screen run",
    )
```

then this would be misleading, right? Maybe to be even more specific, `'protein_search_tool': None,` is not the default, since we read the default from `screen-default-config`. I feel fine about changing the logic less, though!

PR #67, opened 2025-06-25
Benign rebrand

MichaelBarnett:
## Background
This aims to remove the use of the word benign, and replace with the plain English, "low-concern", or "clear" terminology.

## Changes
Any use of the word benign, which makes it to an output or log, has been changed to use low-concern|clear.

### Bug fixes
* 

### Breaking changes 
* The keyword for benign within the yaml has been changed to low_concern.
* The keyword for taxonomy/benign_taxids is changed to low_concern_taxids


MichaelBarnett:
Waiting for changes to Flag, and confirmation of changing all internal variable names specifically for v1.0

alexanian (approved):
one note but LGTM

Issue #63, opened 2025-06-15
Recent pull resulted in dmnd issue


brmagalis:
INFO    │  >> STEP 1: Checking for biorisk genes...
INFO    │ STEP 1 completed at 2025-06-15 18:05:47
INFO    │  >> STEP 2: Checking regulated pathogen proteins...
INFO    │ Processing 5 Diamond dbs using 5 concurrent runs with 8 threads per run.
INFO    │       ERROR: Homology search has failed
ERROR   │ STEP 2: Protein search failed due to an error:
        │  Output of protein taxonomy search could not be processed:
        │ /blue/salemi/brittany.rife/nti/igem_plasmids/output_pAI-10462/pAI-10462.nr.dmnd
INFO    │  Traceback:
        │ Traceback (most recent call last):
        │   File "/blue/salemi/brittany.rife/nti/common-mechanism/commec/screen.py", line 360, in run
        │     self.screen_proteins()
        │     ~~~~~~~~~~~~~~~~~~~~^^
        │   File "/blue/salemi/brittany.rife/nti/common-mechanism/commec/screen.py", line 466, in screen_proteins
        │     raise RuntimeError(
        │         f"Output of protein taxonomy search could not be processed:
        │ {self.database_tools.regulated_protein.out_file}"
        │     )
        │ RuntimeError: Output of protein taxonomy search could not be processed:
        │ /blue/salemi/brittany.rife/nti/igem_plasmids/output_pAI-10462/pAI-10462.nr.dmnd
INFO    │  >> STEP 3: Checking regulated pathogen nucleotides...
INFO    │ No protein hits found, screening entire sequence.
INFO    │       ...no regulated hits
INFO    │ STEP 3 completed at 2025-06-15 18:22:05
INFO    │  >> STEP 4: Checking any pathogen regions for benign components...
INFO    │       ...no regulated regions to clear
INFO    │ STEP 4 completed at 2025-06-15 18:22:05
INFO    │  >> Commec Screen completed at 2025-06-15 18:22:05

alexanian:
In particular, this shouldn't be an "Error" overall or for the `taxonomy_aa` step, since it's okay to have no results.
![Image](https://github.com/user-attachments/assets/c36fbeef-0456-42bf-80a6-9fd3b39bf0ad)

PR #74, opened 2025-07-07
Fixes Diamond empty file outputs.

MichaelBarnett:
## Background
Unlike BlastX, and the other tools, Diamond output files contain no header information. Due to this, the output checks were conflating the logic of "file doesn't exists" with "file exists, but is empty" when determining whether to proceed with Step 2. This was leading to erroneous ERROR statuses being present when Diamond was returning no hits.

**Issues**:
Addresses issue #63 

## Changes
The `check_output()` method is now _only_ used, instead of _as well as_ the `is_empty()` method to determine whether the output of a handler is valid. for further STEP processes. Importantly, Diamond has overridden the method to ignore if the file is empty, but does exist, which is a valid output for Diamond Handlers.

To ensure that the `has_hits`, and `is_empty` is used correctly, the static method status for each has been removed. This has resulted in a refactor of `calculate_noncoding_regions_per_query` to take Query dictionaries and the handler, rather than arbitrary file names.

Furthermore, old unused legacy based code for historic fetching of nc bits has been removed, and the associated test rewritten to incorporate the used methods, as well as correctly skipping a < 50 bp non-coding multi-query output.


commec/tools/fetch_nc_bits.py
```diff
@@ -91,68 +91,11 @@ def _write_nc_sequences(nc_ranges, record, outfile):
...
-        )
-
-    ranges_str = ", ".join(f"{start}-{end}" for start, end in ranges_to_screen)
-    logger.info("Writing noncoding regions [%s] to: %s", ranges_str, outfile)
-    _write_nc_sequences(ranges_to_screen, records[0], outfile)
-
 def _set_no_coding_regions(query : Query):
     query.non_coding_regions.append((1, query.length))
 
 def calculate_noncoding_regions_per_query(
```
alexanian, line 97: sanity-checking:

I think we can also remove `_get_records` and `_write_nc_sequences`, but I wanted to check about the indexing here. So previously we had this in `fetch_nc_bits`:

```python
def _write_nc_sequences(nc_ranges, record, outfile):
    """
    Write a FASTA containing only the parts of the record without a high-quality protein match.
    """
    nc_sequences = []
    for start, stop in nc_ranges:
        # subtract 1 just from `start` to adjust to 0-based index and capture full range
        sequence = record.seq[int(start) - 1 : int(stop)]
        # when parsed from a FASTA file, description includes the id:
        # https://biopython.org/docs/latest/Tutorial/chapter_seq_annot.html#seqrecord-objects-from-fasta-files
        nc_sequences.append(f">{record.description} {start}-{stop}\n{sequence}\n")

    with open(outfile, "w", encoding="utf-8") as output_file:
        output_file.writelines(nc_sequences)
```

and now we have the following in screen_io:

```
        # Generate the non-coding fasta.
        nc_fasta_sequences = ""
        for query in self.queries.values():
            if query.result_handle.status.nucleotide_taxonomy == ScreenStatus.SKIP:
                continue
            nc_fasta_sequences += query.get_non_coding_regions_as_fasta()

        # Skip if there is no non-coding information.
        if nc_fasta_sequences == "":
            logger.info(
                "\t...skipping nucleotide search since no noncoding regions fetched"
            )
            self.reset_query_statuses(ScreenStep.TAXONOMY_NT, ScreenStatus.SKIP)
            return

        # Create a non-coding fasta file.
        with open(self.params.nc_path, "w", encoding="utf-8") as output_file:
            output_file.writelines(nc_fasta_sequences)

```

do we still need to do anything, or is this all handled in `get_non_coding_regions_as_fasta` ?

commec/tools/search_handler.py
```diff
@@ -136,20 +138,22 @@ def _validate_db(self):
                 " File location can be set via --databases option or --config yaml."
             )
 
-    @staticmethod
-    def is_empty(filepath: str) -> bool:
-        """Check if a file is empty or non-existent."""
+    def is_empty(self) -> bool:
```
alexanian, line 141: So I don't think this will work (and I think we're missing a test to catch that this doesn't work) because right now `check_reg_path` has:

```
    if BlastNHandler.is_empty(input_file):
        logger.error(
            "Cannot check for regulated pathogens in empty or non-existent file: %s\n",
            input_file,
        )
        return 1

    if not BlastNHandler.has_hits(input_file):
        logger.info("\t... Skipping regulated pathogens check, no hits in: %s\n", input_file)
        return 0
```

which is using the statcmethod to check the input file

commec/tools/search_handler.py
```diff
@@ -136,20 +138,22 @@ def _validate_db(self):
...
-            return os.path.getsize(os.path.abspath(os.path.expanduser(filepath))) == 0
+            return os.path.getsize(os.path.abspath(os.path.expanduser(self.out_file))) == 0
         except OSError:
             # Errors such as FileNotFoundError considered empty
             return True
 
-    @staticmethod
-    def has_hits(filepath: str) -> bool:
-        """Check if a file has any hits (lines that do not start with '#')."""
+    def output_exists(self) -> bool:
```
alexanian, line 149: propose that we unify all of these as `has_empty_output` / `has_empty_output` / `has_hits` and use a lil convenience method of:

```
    @property
    def _output_path(self):
        """Normalized, absolute path for the output file of this search."""
        return os.path.abspath(os.path.expanduser(self.out_file))
```

(except this doesn't address the `check_reg_path` issue I raised above)

alexanian (changes requested):
Liking the direction but think there are a few tests / etc to tweak; proposed a few renames in https://github.com/ibbis-bio/common-mechanism/compare/diamond-patch...diamond-patch-tessa-patch?expand=1 , feel free to take 'em or leave 'em

commec/tools/search_handler.py
MichaelBarnett: Ah yes, that would be because check_regulated_pathogens() is actually deprecated, and un-used legacy code now. I should just delete it like the old non-coding code. Its still hanging for quick reference. Everything is done in the `update_taxonomic_data_from_database()` method - which is also to be renamed to your suggestion made in the benign rebrand.

commec/tools/fetch_nc_bits.py
MichaelBarnett: 100%, `_get_records`, and `_write_nc_sequences` are now unused. I missed them when deleting the rest of the other legacy code.

commec/tools/search_handler.py
MichaelBarnett: better yet, this is handled in the constructor to a certain extent already, as all these files are passed during object instantiation, we currently only get the absolute path, but it would be better to incorperate os.path.expanduser() too, I think that is a cleaner way to solve the problem - so that any additional implementation changes in the future don't have to remember to call a file sanitization function.

alexanian (approved):

PR #77, opened 2025-07-11
Database validation override

MichaelBarnett:
## Background
Adds a simply database override statement such that if `-R, --resume` is used, and the expected output already exists (i.e. the database handler would not run the search anyway, then the database files themselves are not checked for existence / validity. 

This allows the complete run of a commec screen based on pre-existing output alone - with no large external database file setup [biorisk/nr/nt/benign]. 

This is useful for testing, reproducibility, and tutorial purposes.

Note, we still require correct config yaml imports of the annotations and taxid files for biorisk and benign - these are however very lightweight, and could be bundled with a yaml easily for tutorial output purposes.

### DIFF
--- a/commec/screen.py
+++ b/commec/screen.py
@@ -469,7 +469,7 @@ class Screen:
         """
         logger.debug("\t...running %s", self.params.config["protein_search_tool"])
         self.database_tools.regulated_protein.search()
-        if not self.database_tools.regulated_protein.check_output():
+        if not self.database_tools.regulated_protein.validate_output():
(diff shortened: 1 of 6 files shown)


alexanian (approved):
This makes sense; I made myself confused while reviewing, have now resolved my confusion and written https://github.com/ibbis-bio/common-mechanism/pull/77/commits/5b1f8d20e53e4f3f54bcb88415f015ae51e7e233 to prevent it; merge if you think that's a fine change

PR #85, opened 2025-08-13
Release v1.0.1

MichaelBarnett:
## Changes
* `commec setup` now correctly validates user input (~) for path to
download commec databases into.
* `commec setup` now exits gracefully when stopped with keyboard
interrupts (Ctrl+C for example).
* `commec setup` Includes feedback to the user if a ~name/ user is specifically
invalid, this is to catch ~commec-dbs/ instead of ~/commec-dbs/ typos as
well as username typos.
* `commec screen` DatabaseValidationError when caught, now terminates gracefully.
* Updated example data to match the wiki for v1.0

### Bug fixes
* `commec setup`: User expansion on a path now works correctly
(~/downloadlocation/commec-dbs/ will expand to the user root, instead of
creating a ~ folder in the current directory)
* `commec setup`: The correctly validated path is now carried through to be used when
the folders are actually created, instead of only during the validation
step.


MichaelBarnett:
[conda recipe setup](https://github.com/bioconda/bioconda-recipes/pull/58542), tag made, ready to push into main and update develop.

alexanian (approved):
worked for me!

PR #95, opened 2025-11-07
Release v1.0.2

MichaelBarnett:
## Background
Several small scale bug-fixes will be released as a final version for users to enjoy before having to download updated commec database files for the upcoming Control Lists release.

## Changes
* Fixes erroneous "Matches ." Rationale text that could occur under some Biorisk output conditions.
* Fixed an issue where Benign clears were being applied to all regions of multiple region regulated hit.
* Taxonomic information is sorted in the json output
* A maximum of 10 entries of non-regulated are shown in the json output in the case of a mixed hit result.
* E-values are now included for each individual taxonomic information per hit
* Updated the description for each hit to include a rationale-like output depending on case, rather than just the description of the first entry.
* Rationale texts are added as a column to `commec flag` output csv.
* Additional support for up to 255 characters of filename size, for those abhorrently long filenames.
* More appropriate error handling depending on whether or not --Resume, --Force are used and the presence of pre-existing output files.
* Certain temporary output files - derived from the output of Blastx, Blastn, hmmscan, for example are now kept in the event of a failed run.
* Commec now differentiates between a query having no hits (with all steps being run fully) and no hits, skipped (when taxonomy was skipped for example with --skip-tx) in its output status'

### New features
* Adds additional Taxonomic info for every hit, to aide in human investigation of any particular mixture of regulated and non-regulated results.

### Refactoring
* Minor tidying of the check regulated pathogens function.


alexanian (approved):
As discussed offline, this needed a few fixes for additional edge cases with the "Matches ." but otherwise seems to be working well! Let's release.

PR #96, opened 2025-11-21
Query name fix

MichaelBarnett:
## Background
An issue appeared when trying to run the sample_NIST-screening-dataset-blinded-seqs.fa examples, where the length of the names were too long and were not being processed correctly for internal ID use. This PR addresses that issue and provides a fix.

## Changes
Query names that are not tokenable (i.e. no not have underscores) and are longer than 64 characters are simply truncated to 64 characters.

Raised Errors with the text from same-id queries are correctly passed to the user.

Extended query id from 25 to 64.


alexanian (commented):
This seems like a good hotfix, though, is the reason we shorten the names to 25 characters that it's better for display purposes? I wonder if we should have a `display_name` field and leave `query.name` alone...

I think we took the 25 character limit from [the FASTA spec](https://www.ncbi.nlm.nih.gov/genbank/fastaformat/):

> The SeqID must be unique for each nucleotide sequence and should not contain any spaces. Please limit the SeqID to 25 characters or less. The SeqID can only include letters, digits, hyphens (-), underscores (_), periods (.), colons (:), asterisks (*), and number signs (#). The sequence identifier will be replaced with an Accession number by the database staff when your submission is processed.

but I don't think limit is broadly observed... So we can probably truncate at 64 or something, and then introduce a shorter name for logging?

commec/config/screen_io.py
```diff
@@ -119,14 +119,15 @@ def parse_input_fasta(self) -> dict[str, Query]:
             try:
                 query = Query(record)
                 if query.name in queries:
-                    raise ValueError(f"Duplicate sequence identifier found: {query.name}")
+                    raise ValueError(f"Duplicate sequence identifier generated: \"{query.name}\" from record: {record}\n"
+                                     "Ensure that the first 25 characters for each fasta record are unique.")
```
alexanian, line 123:
```suggestion
                                     "Ensure that the first 64 characters for each fasta record are unique.")
```

alexanian (approved):

PR #97, opened 2025-12-03
Collapse Taxonomic hits based on hit name, rather than the hit accession string.

MichaelBarnett:
The hit name is more likely than accession to not be unique, providing a better user experience for parsing verbose repeated like sequences, so we now collapse hits based on name. 

The functional test outputs were carefully curated to ensure that the result was the same, and required changes to some hit names to ensure that the benign checks treated appropriate regions separately. Also minor updates to the checks for testing regions in screen.

Whilst consider the name or accession as a collapse point, this created complications in checking, as the accession is not typically stored to be checked. In either case, the accession tends to be more unique, and more often the name is shared with unique accessions, rather than the other way around.

### DIFF
--- a/commec/screen.py
+++ b/commec/screen.py
@@ -509,7 +509,7 @@ class Screen:
 
         if exit_status != 0:
             raise RuntimeError(
-                f"Output of protein taxonomy search could not be processed: {self.database_tools.regulated_protein.out_file} {exit_status}"
+                f"Output of protein taxonomy search could not be processed: {self.database_tools.regulated_protein.out_file}"
             )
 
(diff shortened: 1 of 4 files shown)


alexanian (approved):
This is a relatively quick-and-not-too-dirty fix because we had some results that look like this:

Image

Image

and those are silly! Merging on 'subject title' gets us most of the way to collapsing these:

Image

In the future, we should definitely filter out database a bit and we will want to report hits based on control-relevant information (rather than database-subject-relevant information)... but this is good enough for a speedy release!

PR #98, opened 2025-12-04
Update HTML report with summary table and header

alexanian:
## Background
When reviewing some recent user-reported results, I found myself wishing there was a bit more information in the HTML report (including the commec version number and a summmary table). Now this exists!

## Changes

The HTML now features a persistent header containing version information, as well as a summary table showing all queries, with their status and rationale.

image

The results have also been modified to remove the title text from the plotly figure (so it can more easily be copied) and to add a "Jump to top" button.

The main line changes are a large number of additional CSS classes. I thought about doing more changes to how the HTML is generated but wanted to keep this relatively small in scope.


commec/utils/json_html_output.py
```diff
@@ -124,16 +126,45 @@ def generate_html_from_screen_data(input_data : ScreenResult, output_file : str)
...
+        })
+        
+        # Count statuses
+        status_lower = query.status.screen_status.lower()
+        if 'flag' in status_lower:
+            status_counts['flag'] += 1
+        elif 'warn' in status_lower:
+            status_counts['warning'] += 1
+        else:
+            status_counts['pass'] += 1
```
MichaelBarnett: I believe this else statement will count errors, nulls, stops, and skips as passes.

I would recommend creating an array of all screen_statuses, and then map them (either using importance (I notice the int values map nicely into this schema), or a predefined dict) to one of the following short terms: flag, warn, pass, skip, error. And then count the contents for each in the array.

When it comes to displaying status counts, only display skips and errors iff they existed i.e. there were more than 0 in the counts. This could be extended to warnings, but I'd recommend always have at least Pass and Flags, even if there were no flags.

commec/utils/json_html_output.py
```diff
@@ -124,16 +126,45 @@ def generate_html_from_screen_data(input_data : ScreenResult, output_file : str)
...
 
     # Additional template information
     n_query = input_data.query_info.number_of_queries
     plural = "y" if n_query == 1 else "ies"
     html_title = f"Commec Screen Summary: {n_query} Quer{plural}."
+    
+    # Extract basenames for template
+    import os
```
MichaelBarnett: Move to header of file for consistency?

MichaelBarnett (changes requested):
With the exception of other non-warn and non-flag outcomes being interpreted as Passes, looks good to me.

Ran on my own test set data and it looks really swish! Huge improvement over the previous rendition.

commec/utils/json_html_output.py
alexanian: oop yeah, the pain of vibecoding

PR #99, opened 2025-12-05
Species html truncation

MichaelBarnett:
## Background
image

## Changes
image

### New features
* Many species names that are the same are no longer endlessly repeated in the output for html overtext.


alexanian (approved):
Simple change, looks much cleaner

PR #100, opened 2025-12-05
Forces all imported query name related columns to be of datatype str,…

MichaelBarnett:
## Background
There are no protections when import data from blast / cmscan / hmmer to ensure that the datatype for the query name was string, and it may be that query names of a single numerical character i.e.e '0' or '5' are being imported as chars or ints, leading to issues with parsing their respective row data to the query objects imported from fasta.

This might address this by forcing the datatypes on import.


manu-script (approved):
Can confirm that this fixes the issue I encountered with the query names with simple integers on my set up.

alexanian (approved):

PR #102, opened 2025-12-15
Support evalportal format in commec flag output

manu-script:
## Background
The [IBBIS screening evaluation portal](https://evalportal.ibbis.bio/) requires a CSV or TSV with a `UUID` query name column and a `Flag` boolean column where positive inputs are 1 and negative inputs are 0.

### New features
* This PR adds evalportal format support to `commec flag` output through the `-e, --evalportal-format` option. 
* Also added a `test_evalportal_format()` based on the existing `test_flag()`.
 
Note: Uses a "commec_strict" approach following this [paper](https://www.biorxiv.org/content/10.1101/2025.05.30.655379v1):

> To reach their standards, NIST had to use a "commec_strict" approach of counting "warning" as "flag" ― this is because taxonomic hits were getting filtered out by unclassified sequences. We should ignore unclassified sequences and treat "warning + mixed taxonomy" as "flag".

We should relax this approach when the issue of unclassified taxa is fixed.

## NIST Study guide example
Input JSON: [sample_NIST-screening-dataset-blinded-seqs.output.json](https://github.com/user-attachments/files/24164396/sample_NIST-screening-dataset-blinded-seqs.output.json)

Command: `commec flag --evalportal-format <directory containing the .json file>`

Output: [screen_pipeline_status.csv](https://github.com/user-attachments/files/24164425/screen_pipeline_status.csv)

Evaluation Portal result:
image

### DIFF
--- a/commec/flag.py
+++ b/commec/flag.py
@@ -51,6 +51,13 @@ def add_args(parser: argparse.ArgumentParser):
         default=None,
         help="Output directory name (defaults to directory if not provided)",
     )
+    parser.add_argument(
+        "-e",
+        "--evalportal-format",
+        dest="evalportal_format",
+        action="store_true",
+        help="Output format compatible with the IBBIS sceening evlauation portal",
+    )
     return parser
 
 def read_flags_from_json(file_path) -> list[dict[str, str | set[str] | bool]]:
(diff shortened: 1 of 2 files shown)


alexanian (approved):
We both thought about whether there should be a `--fmt` argument instead, but let's just add that if and when we need it, rather than doing so right now?

PR #107, opened 2026-03-25
Remove no-hits warning logic 

manu-script:
## Background
The custom BLAST databases significantly increase the fraction of queries with no hits to the database due to the restricted taxonomic scope around the controlled taxids. Therefore, the "No-Hits Warning" doesn't exclusively signify the in-silico design of the query anymore. This PR removes the no-hits warning logic and makes some other minor changes required for the benchmarking pipeline.

### Breaking changes
* Removed the warning and rationale for queries with "No Hits" and "No Hits (Skipped)". Queries with no hits to the database are now passed as no regions of concern.
### Minor changes
* Removed unused columns in the evalportal format for simplification
* Restricted the Python version to <3.14 for compatibility with Snakemake in the benchmarking pipeline

## Testing
Tested the changes on a small benchmarking test set. The comparison of flags before and after the changes shows all queries with no hits being passed:
image

Before: 
image

After: 
image

### DIFF
--- a/commec/config/result.py
+++ b/commec/config/result.py
@@ -385,11 +385,11 @@ class QueryScreenStatus:
             self.screen_status = ScreenStatus.SKIP
             return
 
-        # If Taxonomy steps were skipped, and we passed, then update to skipped pass.
+        # If Taxonomy steps were skipped, but we passed, then --skip-tx or --skip-nt was used. 
+        # Update to skipped pass.
         if (self.screen_status == ScreenStatus.PASS and 
-                (self.protein_taxonomy == ScreenStatus.SKIP or 
-                self.nucleotide_taxonomy == ScreenStatus.SKIP)
-            ):
+            (self.protein_taxonomy == ScreenStatus.PASS_SKIP_TX or
+             self.nucleotide_taxonomy == ScreenStatus.PASS_SKIP_TX)):
             self.screen_status = ScreenStatus.PASS_SKIP_TX
             return
 
@@ -498,7 +498,7 @@ class QueryResult:
         logger.debug("Updating step status flags for query %s", self.query)
         logger.debug("Current status %s", self.status)
         
-        ignored_status = {ScreenStatus.SKIP, ScreenStatus.ERROR, ScreenStatus.PASS}
+        ignored_status = {ScreenStatus.PASS_SKIP_TX, ScreenStatus.SKIP, ScreenStatus.ERROR, ScreenStatus.PASS}
         
         if self.status.biorisk not in ignored_status:
             self.status.biorisk = ScreenStatus.NULL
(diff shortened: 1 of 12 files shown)


commec/config/result.py
```diff
@@ -578,20 +570,7 @@ def _update_rationale(self,
...
-            state.nucleotide_taxonomy in [ScreenStatus.PASS, ScreenStatus.SKIP] and
-            state.low_concern in [ScreenStatus.PASS, ScreenStatus.SKIP]):
-            # Add an extra caveat if the taxonomy search was skipped
-            if ScreenStatus.SKIP in [state.protein_taxonomy, state.nucleotide_taxonomy]:
-                state.rationale = Rationale.NO_HITS_SKIP_NOTE
-            else:
-                state.rationale = Rationale.NO_HITS
-            return
-
+
```
MichaelBarnett, line 605: Somewhat concerned about the following - previously there was a difference in outcome depending on if skips were stepped, i.e. if --skip-tx was used, and no hits found, the error would reflect that the hits may be found if run in full mode. Likewise, I wonder if there should be an updated rationale text if there were not hits, thus it was a PASS, but we also skipped taxonomy? I may be prudent to identify the pass with a rationale which mentions skipped taxonomy? Perhaps we should go as far as having a PASS_SKIP_TX which represents this case.

Whilst this is somewhat out of scope for this PR, I do think this would be the time to address this change too. I will ping @alexanian about this for her thoughts in this too.

MichaelBarnett (commented):
Overall looks fine, I do want to follow up with discussion about whether want to retain an element of difference in output depending on whether --skip-tx | --skip-nt is used or not.

Won't merge just yet either, as I don't think we want this functionality in with the latest Biorisk DB update - which should be soon.

commec/config/result.py
alexanian: I think the idea of introducing a PASS_SKIP_TX status addresses this nicely (since other outcomes don't really represent a security concern, and we don't want strongly security-logic-relevant stuff to live only in rationale plaintext.)

manu-script:
Agreed. Thanks for making the PASS_SKIP_TX changes @MichaelBarnett. Looks good. We can merge this after the biorisk update release.

manu-script:
@MichaelBarnett PASS_SKIP_TX is being triggered when a protein hit is found, and nucleotide search is skipped. Most likely because the conditions are similar to when --skip-nt is used.

image

MichaelBarnett:
@manu-script, updated which skip's are used during screen when skipped from --skip-tx or --skip-nt settings allowing for PASS_SKIP_TX to be assigned more reliably without having access to the setting when updating a result (by design we want to maintain the update() function to be limited to its own results state)

Have a look, but this seems like a reasonable solution, and it has passed some of

Now:

* SKIP = This step was skipped based on some internal screen situation i.e. no non-coding regions. Or Query too short / long.

* PASS_SKIP_TX = This step was skipped due to --skip-tx or --skip-nt only.

alexanian (approved):
I think this is now ready (once we're ready to release the db)

PR #108, opened 2026-04-01
Low Concern Bugfix

MichaelBarnett:
## Background
Manu came across a critical error in commec screen when testing 100 queries for the upcoming biorisk database release. One of the queries (the final one - although this coincidental) failed an index accession operation for the RNA low concern database after filtering. Interestingly, the bug disappeared when the query is run in isolation.
Further investigations by limiting to the 3 queries present in the RNA low concern outputs revealed that the accession logic on that database for grabbing the top remaining low concern hit was incorrect, accessing the 0th index by name rather than location.

This bug affects non-first queries in multi-query only commec screen runs where the benign steps find multiple outputs such that the index of the affect queries row is not 0.

This was identified on the RNA low concern step, however could also occur in the other low concern steps as well - which share the logic of accession with `[0]`.

## Changes
All low concern steps are now accessed using `iloc[0]`, rather than `[0]`, which ensures the 0th element of the remaining truncated benign data (i.e. the topmost, truncated benign hit) is reported on. 

### Bug fixes
commec screen no longer crashes in multi-query runs where a two or more queries have benign hits.

### DIFF
--- a/commec/screeners/check_low_concern.py
+++ b/commec/screeners/check_low_concern.py
@@ -81,13 +81,13 @@ def _filter_low_concern_proteins(query : Query,
         return []
 
     # Report top hit for Protein / RNA / Synbio
-    low_concern_hit = low_concern_protein_for_query_trimmed["subject title"][0]
+    low_concern_hit = low_concern_protein_for_query_trimmed["subject title"].iloc[0]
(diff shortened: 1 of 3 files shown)


manu-script (approved):
Can confirm this fixes the issue.

PR #109, opened 2026-04-13
Merge high-identity hits when calculating noncoding regions

alexanian:
## Background
We got a bug report from a user as follows:
image

The "coding region" was being registered as ending at the end of the encompassed hit, instead of the union of the two high-identity hits.

Previously _get_ranges_with_no_hits walked hits pairwise in sorted-by-start order and used hit_ranges[-1][1] (= 600, the end of 400, 600]) as the rightmost covered base, incorrectly returning [(1, 199), (601, 1000)] and re-screening [601, 800] at the nucleotide level even though it is covered by the [200, 800] protein hit.

### Bug fixes
* Update calculation of coding regions to correctly return ranges
* Add regression test for bug

### DIFF
--- a/commec/tests/test_fetch_nc_bits.py
+++ b/commec/tests/test_fetch_nc_bits.py
@@ -19,36 +19,53 @@ from commec.tools.blastx import BlastXHandler
 DATABASE_DIRECTORY = os.path.join(os.path.dirname(__file__), "test_dbs")
 
 @pytest.mark.parametrize(
-    "hits, nc_ranges",
+    "hits, query_length, nc_ranges",
     [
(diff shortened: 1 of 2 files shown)


MichaelBarnett (approved):
Adds a refactor to the fetch nc bits tests to take query length, and adds tests for overlaps.

Creates a merged representation of the unique hit_ranges, by checking the last member of the merged list, and updating its end point with the new end point, rather than adding new start-end pairs if the new start was smaller than the end point (i.e. overlapping).

This merged list (instead of hit_ranges) is then used for finding the non-coding regions > 50 bps.overlapping.

Minor re-arrangement of code constants e.g. query_length.

Looks good to me.

PR #110, opened 2026-04-15
Release v1.0.4

MichaelBarnett:
## Background
Minor updates for `commec screen`, including bug-fixes, coupled with an updated Biorisk and low-concern databases (databases v.1.0.1).

## Changes
* benign database files are expected to be named low_concern.* by default, as per the v1.0.1 databases release.
* E-Value cut-offs are modified dependent on query sequence length for improved sensitivity.
* Non-coding region calculations no longer truncate when protein hits overlapped under some circumstances.
* Custom base_paths in the configuration yaml that are not the default are correctly propagated throughout the input settings. 

### Breaking changes 
* Default yaml config files will look for low_concern.* instead of benign.* database files within the low_concern directory.

### DIFF
--- a/commec/config/constants.py
+++ b/commec/config/constants.py
@@ -4,6 +4,14 @@
 # SCREENING
 MINIMUM_QUERY_LENGTH = 41
 
+# BIORISK E-VALUE FILTERING
+# Sequences shorter than this threshold use a length-dependent E-value cutoff.
+BIORISK_SHORT_QUERY_NT_THRESHOLD = 200
+# Exponent for the length-dependent E-value formula: E < 1 / (1 + L^EXPONENT)
+BIORISK_SHORT_QUERY_EVALUE_EXPONENT = 2.598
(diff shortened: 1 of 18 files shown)


MichaelBarnett:
Notes on this:
* Protein Tx hits are sorted on % identity, not E-value, thus we see "overlapping" smaller hits of protein - this may be considered outside the scope of this update.

* Based on tests, very much improved nucleotide tx outputs, due to the bugfix in non-coding region overlapping coordinate issue.

* Biorisk database is working and catching new hits previously missed with the updated e-value and biorisk profiles.

* Minor "benign" word removal that sneaked into the html.

It would be nice to get this out asap to remove further benign vs low_concern issues with yamls. I will explicitly ensure that this change is mentioned foremost in the release notes created to ensure people get a heads-up after updating.

manu-script (approved):
Looks good. We can release this in sync with the biorisk database updates.

PR #111, opened 2026-04-20
Search parameter optimizations to improve performance

manu-script:
## Background
This PR fine-tunes the parameters of search tools to improve speed and sensitivity. It is based on some observations: 
- The CPU thread usage of HMMER, CMSCAN, and BLASTN often saturates quickly due to I/O overhead. 
- The default task parameter of BLASTN is not suited to detect diverged/engineered non-coding regions.
- BLAST v2.17 introduced a faster protein search task. 

## Changes
- **HMMER/CMSCAN thread cap**: Added `MAX_THREADS = 4` class constant to both `HmmerHandler` and `CmscanHandler` (enforced via a new `_apply_thread_cap` hook in the base `SearchHandler`) to prevent I/O saturation at high thread counts. If the user requests more than 4 threads, it is capped with a warning log.
image

- **BLAST parameter tuning**: 
    - Switched BLASTN task to `dc-megablast` with non-coding template settings, e-value tightened from `10` → `1e-5`, `max_target_seqs` reset to default `500`, `culling_limit` decreased from `5` → `1`, and `-mt_mode 1` to split work by batched query rather than by database, improving I/O performance under multithreading. 
    - Switched BLASTX task to `blastx-fast`, removed alignment params with default values to avoid confusion, and reduced `max_target_seqs` from `5000` → `500` and `culling_limit` decreased from `5` → `1`. 

## Benchmarking
Testing these changes on the NIST, SBRC, and iGEM test sets showed further speed gains with the reduced BLAST databases. It also slightly improved the sensitivity metrics, which pushed the NIST March evaluation dataset over the passing threshold.


commec/tools/cmscan.py
```diff
@@ -15,6 +15,8 @@
 class CmscanHandler(SearchHandler):
     """A Database handler specifically for use with Hmmer files for commec screening."""
 
+    MAX_THREADS = 4
```
MichaelBarnett: This all looks well and good - however, do we want to hard-code in the maximum number of threads. Can we imagine situations where the i/o bound nature of hmmscan isn't an issue - for example in HPC situations?

Understandably this is likely not a big issue for our users but I think we should retain the ability to change this at will still.

I want to suggest either:

* modifying the default yaml config to include a max_threads for either the specifically affected steps - but perhaps its more prudent to just give them all such treatment, which simply gets passed to the constructor or modified after instantiation.

* Put a HMMSCAN_MAX_THREAD_LIMIT , and CMSCAN_MAX_THREAD_LIMIT constants into the constant.py, and read them here instead of 4, such that at least the constants are in a single place for those who wish to modify the source code.

MichaelBarnett (changes requested):
Overall seems fine, but slight concern at where the hard coding of MAX_THREADS occurs, and that its also not easily modifyable.

I am leaning towards leaving the constant in constants.py - rather than a more verbose settings change. As that is where we want to put "Easily modified, but you really shouldn't unless you really know what you are doing" constants.

MichaelBarnett (approved):
Nice, nit-picky on my part but it is nicer to have these in one place, and clearly labelled in purpose!

manu-script:
Thanks for pointing this out! Constants.py is definitely the right place to define this.

Also, this issue is documented in the HMMER manual. Just putting it here as a rationale for specifically choosing 4 threads after testing on HPC and SSD setups for future reference.
image

commec/tools/blastx.py
```diff
@@ -23,17 +23,12 @@ def __init__(
         super().__init__(database_file, input_file, out_file, **kwargs)
         # We fill this with defaults, however they can always be overridden before screening.
         self.arguments_dictionary = {
+            "-task": "blastx-fast",
             "-num_threads": self.threads,
+            "-mt_mode": 1,
             "-evalue": 1e-10,
-            "-word_size": 6,
```
alexanian, line 28: Just realizing that I think we need to restore this smaller word size - we're going to lose some smaller 50bp-ish matches if we stick to the default word size of 11

PR #112, opened 2026-04-22
Query description handling

manu-script:
## Background
The sequence descriptions in the FASTA header are currently being concatenated with the seq ID while replacing all whitespaces and illegal characters with underscores. The query name is then constructed from the first 64 characters and propagated to the HTML, JSON, and CSV outputs. Conventionally, the query ID (up to the first space in the header) and the description following the ID should be reported to allow matching of original seq IDs for downstream analysis (e.g., benchmarking). This PR modifies the query name and description handling to support this. 

## Changes
* The FASTA header lines are now exempt from replacing whitespaces in the FASTA cleaning step, which allows correct parsing of the header. Added description attributes to Query and QueryResult classes. The original seq IDs and descriptions are now reported in the JSON, HTML report, and CSV output. 
### Bug fixes
* Fixed a likely bug that was overwriting original Query seq_record.id by modifying the same SeqIO object after query construction.

## Example
```
>sequence_ID this is the description
ATCGTACATCGTAGCATGCTAGCTAGCTAGCTAGCTAGATCGATCGATGCATGCATGCTAGCTAGCTAGCTAGCTAGCTAGCTACGATGC
```
Before:
image
After:
image

### DIFF
--- a/commec/config/query.py
+++ b/commec/config/query.py
@@ -21,6 +21,7 @@ class Query:
         Query.validate_sequence_record(seq_record)
         self._seq_record = seq_record
         self.name = self.create_id(seq_record.id)
+        self.description = seq_record.description[len(seq_record.id):].strip()
         self.non_coding_regions : list[tuple[int, int]] = [] # 1 based coordinates for Non-Coding Regions.
         self.result : QueryResult = None
         self.translations: list[QueryTranslation] = []
--- a/commec/config/result.py
+++ b/commec/config/result.py
@@ -421,6 +421,7 @@ class QueryResult:
     Container to hold screening result data pertinant to a single Query
     """
     query: str = ""
+    description: str = ""
     length: int = 0
     status: QueryScreenStatus = field(default_factory=QueryScreenStatus)
     hits: dict[str, HitResult] = field(default_factory=dict)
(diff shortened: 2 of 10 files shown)


commec/config/query.py
```diff
@@ -21,6 +21,7 @@ def __init__(self, seq_record: SeqRecord):
         Query.validate_sequence_record(seq_record)
         self._seq_record = seq_record
         self.name = self.create_id(seq_record.id)
```
MichaelBarnett, line 23: Initially custom parsing of SeqID's into a single "id" regardless of description was done to both support previously weird custom fasta's we had been using (i.e. if you used a cleaned fasta as input to commec screen haha circular problems),as well as protect against overly large ID strings during processing, but mainly for users to have non-standard fasta headings if they happened to.

Standardising our approach with what is accepted as a Fasta was discussed, but never fully implemented (i.e. hence the previous 25 char limit - which is literally just grabbing the ID which is that long, and this also guaranteed uniqueness, even for duplicate ID's with different descriptions which may occur in some test sets out of laziness etc)

That all being said, I am happy with the idea to switch to the formal method of fasta interpretation and separation of the description whilst retaining it as a new field in the output jsons - which is what this PR achieves.

The loss of still being able to parse some "arbitrarily formed" fasta headings is probably not a big issue, and people should really be using fasta headings that adhere to the format.
We still are raising the duplicate ID error, so ultimately this puts that onus on the user.

Importantly, the id used to map the query to the json outputs, and the id sent to the new fasta is the still the same.
 
I'll divert to @alexanian for an additional thought on these points, as I know we have run into plenty of gatchas dealing with fasta ingestion, but I suspect this continues to be a move in the right direction.

MichaelBarnett (approved):

PR #115, opened 2026-05-11
Skip very long queries

manu-script:
## Background
While screening the iGEM registry sequences, HMMER crashed on encountering a very long sequence. There seems to be a hard limit of 100K for sequence length in hmmscan. This PR adds a maximum query length constraint similar to the existing minimum query length constraint.

image

## Changes
* Added a MAXIMUM_QUERY_LENGTH constant and updated the query length checking logic to skip sequences longer than this limit, and changed the rationale for skipped sequences to reflect length constraints.


commec/config/result.py
```diff
@@ -292,7 +292,7 @@ class Rationale(StrEnum):
     NO_HITS = ("No matches found during any stage of analysis. "
                 "Sequence risk is unknown, possibly generated in silico. ")
     NO_HITS_SKIP_NOTE = NO_HITS + "Matches may be found if re-run without skipping steps."
-    TOO_SHORT = "Query is too short, and was skipped."
+    LENGTH_OUT_OF_RANGE = "Query is either too short or too long, and was skipped."
```
alexanian: I would prefer that we're specific here ― we should say whether it was too long or too short (and ideally would include our length limits here)

commec/screen.py
```diff
@@ -332,12 +332,17 @@ def setup(self, args: argparse.Namespace):
...
 
-                # Determine short querys as skipped:
+                # Determine out-of-range queries as skipped:
                 if query.length < MINIMUM_QUERY_LENGTH:
                     logger.debug("%s length %i is less than %i",
                                     query.name, query.length, MINIMUM_QUERY_LENGTH)
                     qr.skip()
                     continue
+                elif query.length > MAXIMUM_QUERY_LENGTH:
+                    logger.debug("%s length %i exceeds maximum %i",
```
alexanian: I wonder if these should be logger.warning instead, I feel like users might want to know this (in addition to developers).

Image

alexanian (changes requested):

commec/screen.py
manu-script: Yeah, that's better. Changed it to `logger.warning` in 5022311

commec/config/result.py
manu-script: I've added more specific reasoning to the rationale to reflect this in 5022311. 

image

commec/config/result.py
```diff
@@ -571,7 +571,7 @@ def _update_rationale(self,
             return
 
         if state.screen_status == ScreenStatus.SKIP:
-            state.rationale = Rationale.TOO_SHORT
+            state.rationale = f"{Rationale.SKIPPED} {self._skip_reason}"
```
MichaelBarnett: This has technically violated the idea of "a result object should be able to determine its own state and rationale at all times", separation of concerns design principle we have been using.

By passing in the skip reasoning str during screening, we would technically loose that information if we imported a raw json for any other reason (think restarting a screen run, ingesting json for html outputs, or creating valid json outputs during crashes or early exits).

I think the separation of concerns here is important, and the json data should be self sufficient.

The concept is when a result object calls .update() on itself, it has all the necessary state it needs to complete its own rationales etc.

Currently we got away with this as the only "reason" to skip biorisk was a short query, and we could use that as an indicator to determine the rationale.

I think the solution is to have a custom skipped screen statuses for too_short and too_long. And store those in the screen_status, whilst keeping all the other steps simply skipped, and calling different versions of skip() and tying that to the rationale is the best way forward.
I've made some changes to that effect in commit e9bc49a.

(Note, In the past we had mucked this up with the no_hits_warning logic, and I wasn't too happy about it with myself then either, but its much better now its been removed! =D.)

MichaelBarnett (changes requested):
Consider cherry-picking e9bc49a to keep the state of result the sole source of info in determining rationale logic.

commec/config/result.py
manu-script: Thanks for the suggestion. This is an architecturally better solution. I wasn't happy with the skip "reason" string being passed either. I've merged the edits from e9bc49a.

PR #116, opened 2026-05-13
Validate BLAST/Diamond databases by exact prefix

alexanian:
I owned myself by entering in a database name wrong, and it didn't get caught until the actual BLASTN tool call, and I was sad.

## Background
The previous BlastHandler._validate_db globbed for `*<basename>*` in the database directory, which falsely matched when a user-configured prefix was a substring of the actual one on disk (e.g. configured `nt` against `core_nt.nhr`). The mismatch made it past validation and surfaced later as a cryptic "No alias or index file found" error from BLAST itself.

## Changes
Split validation per handler:
  - BlastNHandler checks <prefix>.nhr / .nal / .<N>.nhr
  - BlastXHandler checks <prefix>.phr / .pal / .<N>.phr
  - DiamondHandler checks <prefix>.dmnd / .<N>.dmnd (strips trailing .dmnd from configured path for prefix normalisation)

We should sanity-check that we'll never have BLAST / DMND databases without these prefixes, but I think we won't?

No index file found (because the index files were /core_nt/core_nt, not /core_nt/nt)
```
  BLAST Database error: No alias or index file found for nucleotide database
  [/capstor/store/cscs/2go/go074/databases/core_nt/nt] in search path                   
  [/capstor/scratch/cscs/talexani::]. Please verify the spelling of the BLAST database
  and its molecule type.                                                                
```


MichaelBarnett (approved):
Exactly why we have our handler abstractions. This is a good change to be specific about how we handle our actual database file detections. I think its a safe assumption that the databases have the given file suffixes expected. If this is bad assumption in the future it will be relatively minor to fix, and also touched empty files will suffice as a temporary work-around.

PR #117, opened 2026-05-14
Be less clever with the config YAML, and simply throw more errors

alexanian:
## Background

I frustrated myself when getting set up in a new environment due to a few places where I think we're a bit too clever with the configuration handling, and it leads to swallowing misconfiguration and confusing error messages. 

Specifically, I passed `-y config.yaml` with a valid relative path, which didn't resolve in the job's working dir, then fell back to package default, and then gave me the error `Screening database directory not found at: /scratch/me/commec-dbs/biorisk`, and I was like "yeah, I know?? I gave you a different path??"

## Changes

1. **Raise `FileNotFoundError` when `--config` YAML path doesn't exist**.
2. **Reject unknown keys in user-provided YAML** — typos like `databse:` were         
  previously dropped with only a `logger.warning`; now they raise. Maybe this was wrong? Warning was a deliberate choice but I currently feel like it was the wrong one.
3. **Require explicit `base_paths.default`; enforce absolute paths in YAML**. Do not include the relative `commec-dbs/` default in the packaged YAML, why would we know where it's installed on a user's machine. Then raise `IoValidationError` if `base_paths.default` isn't set, pointing at the three ways to set it (`-d`, `-y`, or `commec setup`).
  * Also, perhaps more controversially, reject relative paths anywhere in user YAML (ambiguous: relative to CWD? to the YAML file? to install?). CLI `-d` may still be relative and is normalized via `expand_and_normalize`.  
  * INFO-log when `-d` shadows a YAML-supplied base.
4. **Defensive raise in `SearchHandler` on non-absolute database path**:  If any future code path bypasses `ScreenIO` validation, it now fails  at the handler boundary instead of silently resolving against CWD.
5. **Stop mutating packaged YAML in `commec setup`** — `update_default_db_base_path`   used to write the install dir into the packaged YAML. Stateful package mutation breaks across reinstalls, containers, and editable installs. Replaced with `emit_config_snippet`, which prints a YAML snippet and interactively offers to save it for the user to pass via `-y`.
   * @MichaelBarnett you might want to modify this, I feel like you had thoughts on how we could be helping users set config in setup.

### Breaking changes

Existing setups that relied on the packaged `base_paths.default` will now error at startup. Two ways to migrate:

- Re-run `commec setup` and save the emitted snippet, then `commec screen -y <path>`, 
- Pass `-d /path/to/dbs` on the CLI. 

The error message points at both options.
                                                                                        
### Tests recommended (by the AI lol)
  - [ ] Run with no `-d`, no `-y`, no `commec setup` → error names all three ways to set
   `base_paths.default`.                                                                
  - [ ] `-y /missing/path.yaml` → `FileNotFoundError` immediately.
  - [ ] YAML with typo'd key → `IoValidationError` listing the key.                     
  - [ ] YAML with relative path → raises with "must be an absolute path".               
  - [ ] `-d ./dbs` (relative) → expands to absolute, runs.                              
  - [ ] `-d /other` overriding YAML base → INFO log "overriding YAML                    
  base_paths.default".                                                                  
  - [ ] Reproduce original bug: invalid `-y` path now errors pointing at the YAML, not  
  at a spurious database path.


commec/setup.py
```diff
@@ -646,9 +643,9 @@ def do_setup(self):
 
             os.remove(filename_zipped)
 
-        # Default config file installed with commec package should by point to the newly
-        # installed databases
-        self.update_default_db_base_path(DEFAULT_CONFIG_YAML_PATH, self.database_directory)
+        # Emit a config snippet pointing at the freshly installed databases and offer
+        # to save it. The user passes the saved file with `commec screen -y <path>`.
+        self.emit_config_snippet(self.database_directory)
```
MichaelBarnett, line 648: doing this here breaks the --auto cli option, as it expects a user input, which would pause automated install runs.

Although I like the idea of this occuring _after_ successful database download and setup. It probably belongs during the initial setup phases. 

I've written up some changes in 6e0d879 which can be cherry picked.

MichaelBarnett (changes requested):
Just needs to have the --auto flag cli behaviour addressed. But otherwise looks good to me.

AgentK9:
Some thoughts I'm dropping here - more related to `commec setup` rather than configuration.

1. `commec setup` has only 2 flags - what if I want to set the directory or change the download urls without running interactively? (I've had to do this multiple times for tools I run at work, so this is almost certainly a feature users want).
2. `commec setup` does not currently check whether `commec-dbs` (or the specified folder) actually has anything in it before downloading - it would be great to check if the databases were previously downloaded and still the latest first, rather than blindly downloading. I know that the blast database downloads are generally a bit smarter, but still applicable. This could also be used for better error handling for `commec screen` - instead of just running `commec setup`, say `commec setup --repair` or something to specifically check what does and doesn't exist and auto-download. `-a` is still applicable - it would bypass all of these arguments.

I'd also like to recommend (though I understand if this is shot down) using a library like [`click`](https://click.palletsprojects.com/en/stable/) or [`typer`](https://typer.tiangolo.com) for better developer experience and less code.

I can create issues for these if you would prefer!

MichaelBarnett:
@AgentK9 commec setup is going to undergo some pretty major changes as we switch to the R2 bucket databases and versioning system, which I believe will address both of these issues.

AgentK9:
I scaffolded out the CLI with typer on my branch here: https://github.com/AgentK9/common-mechanism/tree/various-improvements

PR #118, opened 2026-06-04
Dynamically update stale regulated taxids

manu-script:
## Background
We currently compare the taxids of BLAST hits against a static list of regulated taxids. NCBI taxonomy updates to taxids can break this logic. For instance, the taxid for Orthoebolavirus genus has been recently updated to 3044781, but we use 186536 in our reg_taxids list. We also have 186538 for the Zaire ebolavirus strain in our list, which is still valid. Due to this, we flag any hits to ZEBOV, but pass hits to any other strain within the genus as non-regulated taxa, while using the latest NCBI taxonomy database. 

image

As different versions of the NCBI taxonomy database may use different taxids for some regulated taxa, we should dynamically update the reg_taxids from the version that was used for screening.

## Changes

- Added a function to retrieve the current canonical taxids from the taxonomy database used for screening and update the reg_taxids and vax_taxids.

### Bug fixes
- Fixed an issue where SoCs were not being flagged due to updated taxids

Before:
image

After:
image

### DIFF
--- a/commec/screeners/check_reg_path.py
+++ b/commec/screeners/check_reg_path.py
@@ -13,7 +13,7 @@ import os
 from dataclasses import asdict
 
 import pandas as pd
-
+import pytaxonkit
 from commec.tools.search_handler import SearchHandler
 from commec.config.query import Query
 from commec.tools.blast_tools import (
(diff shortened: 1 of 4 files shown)


alexanian:
We should make sure to add this specific mapping to  `historic_taxids`: https://github.com/ibbis-bio/control-lists/

I wonder if @MichaelBarnett we should push an update to reg_taxids that represents `--all` (i.e. only taxids in multiple jurisdictions), just so we update this before dynannos are fully ready.

MichaelBarnett:
deprecated taxids, as found in merged.dmp, are already accounted for in the dynamic_annotations and control_list generation workflow. (It however assumes that the latest taxonomy db is used when generating control lists - which _should_ be true, but doesn't address updates to taxonomy dbs when we havent updated the control list.)

By checking things at run-time, we actually go back to being depending on requiring Taxonomy dbs - this was to be removed in Dyna-anno, as we almost have our own "custom controlled taxonomy db" in effect from that.

In its current form this PR seems to update all our regulated taxids to their modern merged versions (which is good), and compares them to the modern merged versions of whatever taxids the blast database had - this elegantly ensures we are always comparing like-with-like in the form of correct/most updated taxid.

I wonder if we revisit the idea in Dynamic_annotations for the removal of the TaxonomyDB search. After working with it more myself, perhaps it is okay to put the onus of having an up to date taxonomyDB on the user. It is reasonably fast to use and is only ~300 mbs, and the main speed issues we were concerned about I think were conflated with the trim overlaps function ... either way, we could also run this on every taxid imported via control_lists, we could also make including a taxonomydb optional - and run the update if it was provided.

Regarding updating reg_taxids with --all, I think this is a bad idea, as there are too many specific weird instances that aren't really controlled by a biological safety point and more export controls - think of the many items in the Brazil control lists for example which simply don't apply to the rest of the world. We could perhaps generate a list using the CCL, UK, China, and US lists - they have considerable more overlap and any differences are more aligned to safety specifically?

MichaelBarnett:
Regarding the tests failing - looks like the test environment isn't setting up a pseudo taxonomydb file structure to use, will need to mock in a replacement.

MichaelBarnett (approved):
lgtm, we're going to cut a quick release for this to ensure that the most recent commec doesn't have these issues.

PR #122, opened 2026-06-15
Feat/parallel blastdb decompress

saarantras:
This is a small change which parallelizes the blast_nt and blast_nr unzip stage of commec setup, improving setup time.

### DIFF
--- a/commec/setup.py
+++ b/commec/setup.py
@@ -40,6 +40,9 @@ def parallel_decompress(directory, jobs=None):
     so we extract them in parallel across ``jobs`` cores (default: all cores).
     """
     jobs = jobs or os.cpu_count() or 1
+    # Resolve to an absolute path so the ``tar`` calls below (run with
+    # ``cwd=directory``) cannot double-prefix a relative directory.
+    directory = os.path.abspath(directory)
     tarballs = glob.glob(os.path.join(directory, "*.tar.gz"))
     if not tarballs:
         return
@@ -100,7 +103,10 @@ class CliSetup:
         Once Start is successfully returned, the setup process can begin.
     """
 
-    def __init__(self, automate: bool = False):
+    def __init__(self, automate: bool = False, jobs: int | None = None):
+
+        # Parallelism for decompressing downloaded BLAST volumes.
+        self.n_jobs = jobs or os.cpu_count() or 1
 
         self.latest_version = get_latest_commec_database_release_tag()[0]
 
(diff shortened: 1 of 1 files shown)


commec/setup.py
```diff
@@ -29,6 +31,48 @@
 C_B_BLUE = "\033[48;5;17m" # Colour Background Blue
 C_RESET = "\033[0m" # Reset Console Formatting.
 
+
+def parallel_decompress(directory, jobs=None):
```
MichaelBarnett, line 35: This is a good idea, and looks like a perfectly fine implementation.

I do wonder if we should add in a way for the user to control the number of threads if desired. Likely best achieved with an additional and optional CLI argument, which is stored in the CliSetup state `self.n_jobs = arg or os.cpu_count() or 1`  and passed around as appropriate?

MichaelBarnett (changes requested):
Still need to run an actual test, but is a good feature.
Minor update to include user cli control of number of desired threads is probably a good nudge to add to the scope for this PR.

commec/setup.py
saarantras: Good idea. Added as you described, except that I defaulted to max parallelism. (Not sure if this is ideal?) Restricted to just unzip parallelism, left downloads alone, though those could theoretically be parallelized (ignored since I expect they will be soon superseded by manu's small database & NCBI might be mad if we hit them with too many requests anyway). I ran a test (download fresh + screen) and it looks good, you might want to try it yourself to check me.

MichaelBarnett: 100% Commec setup is about to get an overhaul which I am currently working on which will simply download necessary databases from the R2 bucket, as part of our new database versioning system. Regardless, unzipping in parallel will make sense there too I suspect. (Although overall the time of all of these steps will be drastically reduced based on the nature of smaller files).

I am skeptical of the returns for parallel downloading (downloading 2 things tends to download both at half speed unless your on some sort of hpc) - but might be something we could consider for the R2 fetching?

MichaelBarnett:
Just a note, we Squash merge features into "develop" branch, and only merge into main for a release.

commec/setup.py
saarantras: If the bottleneck on donwnload time is local download speed, then parallelism doesn't help. The alternative, which is "ncbi sets a per download speed cap". I think you're right you'd normally hit the first before the second, so parallelism there probably wouldn't help. Given the upcoming refactor, maybe i'll just leave this as a "convenience" feature for me as i do testing, and it can be deleted or rewritten after you've switched to the R2.

MichaelBarnett (approved):
Looks better - I am about to rewrite all of setup, so will use some of this logic to keep threading compatibility. Please squash merge into Develop, not main branch.

MichaelBarnett:
Due to the setup update, and tiny new databases, this has become somewhat redundant, although I suppose it could be fun to parallelize that too. I will close this for now, but we can hold onto the branch for when we need the parallel code.

PR #123, opened 2026-06-18
SAST Findings

AgentK9:
## Background

This is one of the steps to improve the generic cyber stance of commec, and has the side effects of ensuring that it's easier to maintain and understand the code.

Essentially, I installed [ruff](https://docs.astral.sh/ruff) and ran `ruff check --fix`, `ruff check --select I --fix`, and `ruff format`. These three commands find and fix common errors, ensure a consistent format and order for imports, and format the code.

I also commented out a line in [json_html_output.py](commec/utils/json_html_output.py) to remove a dead code finding from `ruff check`.

There are 2 follow-on steps from this:

1. Add ruff to CI. GitHub Actions doesn't reallllyyyy lend itself well to using ruff, but I can add it to a new workflow that's "linting" or something.
2. Run [ty](https://docs.astral.sh/ty) on this codebase. Essentially, that will be type checking, and will likely result in more impactful changes.

Also note that this will improve collaboration, since this sets the standard for what code should look like, and will prevent some common formatting issues in PRs.

**Issues**:

### Refactoring
* Ran `ruff format`, `ruff check --select I --fix`, and `ruff check --fix`
* Commented out `r, g, b` line in `update_layout` since it was not used (found by ruff)
* Added ruff configuration to silence some errors, ensure that only relevant code was checked (note - I ran `ruff format` before adding this configuration, so dev_scripts/ and tests/ were both changed - I can revert if preferred).

### DIFF
--- a/.devcontainer/Containerfile
+++ b/.devcontainer/Containerfile
@@ -1,0 +1,14 @@
+FROM mcr.microsoft.com/devcontainers/miniconda
+USER root
+
+COPY environment.yaml /requirements/
+
+# Drop the editable install of this project (the source isn't in the build
+# context); it's installed post-create once the workspace is mounted.
+RUN grep -v -- '-e \.' /requirements/environment.yaml \
+    | grep -v -- '- pip:' > /requirements/environment.build.yaml \
+    && conda env create -f /requirements/environment.build.yaml
+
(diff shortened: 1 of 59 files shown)


AgentK9:
Will run these commands after notable changes are merged

PR #124, opened 2026-06-24
Switched blast mode to 0

saarantras:
Letting blast pick how it wants to use its cores seems to increase utilization & performance for both large and small queries (tested on 12 threads).

https://www.ncbi.nlm.nih.gov/books/NBK571452/

| query | seqs | mt_mode | total (s) | blastx nr (s) | blastx cores | blastn nt (s) | blastn cores |
|---|---:|---:|---:|---:|---:|---:|---:|
| builtin | 5 | 0 | 499 | 254 | 11.4 | 239 | 10.8 |
| builtin | 5 | 1 | 2667 | 1496 | 2.1 | 1164 | 2.1 |
| bronze | 876 | 0 | 5151 | 4232 | 11.9 | 502 | 11.5 |
| bronze | 876 | 1 | 9513 | 6172 | 10.3 | 2923 | 2.0 |

### DIFF
--- a/commec/config/screen_tools.py
+++ b/commec/config/screen_tools.py
@@ -59,6 +59,7 @@ class ScreenTools:
                     input_file=params.nt_path,
                     out_file=f"{params.output_prefix}.nr.blastx",
                     threads=params.config["threads"],
+                    mt_mode=params.config["mt_mode"],
                     force=params.config["force"],
                 )
             elif params.config["protein_search_tool"] in ("nr.dmnd", "diamond"):
@@ -84,6 +85,7 @@ class ScreenTools:
                 input_file=params.nc_path,
                 out_file=f"{params.output_prefix}.nt.blastn",
                 threads=params.config["threads"],
+                mt_mode=params.config["mt_mode"],
                 force=params.config["force"],
             )
 
@@ -100,6 +102,7 @@ class ScreenTools:
                 input_file=params.nt_path,
                 out_file=f"{params.output_prefix}.low_concern.blastn",
                 threads=params.config["threads"],
+                mt_mode=params.config["mt_mode"],
                 force=params.config["force"],
             )
             self.low_concern_cmscan = CmscanHandler(
--- a/commec/screen-default-config.yaml
+++ b/commec/screen-default-config.yaml
@@ -24,6 +24,7 @@ databases:
   taxonomy:
       path: "{default}taxonomy/"
 threads: 1
+mt_mode: 1
 diamond_jobs: null
 do_cleanup: False
 force: False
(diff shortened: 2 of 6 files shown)


manu-script:
Was this tested on the full nr & core_nt dbs or the beta version of small dbs?

saarantras:
> Was this tested on the full nr & core_nt dbs or the beta version of small dbs?

Beta small!

manu-script:
Were the DBs stored on fast SSD/NVMe? I think the results were different on slow NAS on HPC (256 threads).

saarantras:
> Were the DBs stored on fast SSD/NVMe? I think the results were different on slow NAS on HPC (256 threads).

Fast SSD. I will run a quick test on slow HDD (aws has nas, so I can try to approximately replicate those conditions) and append the results here (it will probably take a little while to run).

saarantras:
@manu-script Good catch, it makes a HUGE difference. mt_mode 1 is way, way better in the slow HDD regime :

| step | mode 0  | mode 1  | mode 1 speedup |
|---|---:|---:|---:|
| blastx (nr) | 13h 59m | 1h 17m | ~11x |
| blastn (nt) | ~10h (killed >7.5h) | 2h 05m | -- |
| **total** | **~24h** | **3h 22m** | -- |

 (only tested with 12 threads on an aws st1, and note that some tests I cancelled before they finished: already running way too long).

I've updated the PR to instead add a command-line arg where you can pass mt_mode. It defaults to 1, so the previous behavior stays the same, but adds the option for other settings, so I can set to 0 for commec-in-a-box where i get performance gains.

manu-script:
Yeah, I figured this would be the case.` mt_mode 1` is almost always better with slow DB access, but `mt_mode 0/2` might perform better if there is fast DB access. Having the ability to switch mt_mode in different SSD/HDD setups is useful, but I'm not absolutely sure if introducing a new CLI arg is the best way (given that it is specific to blast and may become obsolete if we switch to a kmer-based tool in the future).

Pinging @MichaelBarnett to weigh in here.

saarantras:
I'll just set it in a yaml downstream to avoid argument cruft then.

(On the commec-in-a-box hardware (fast SSD) I've been running with `0` since I can test stuff faster / will be much faster for end users.)

Also if I understand correctly (which i might not), mt_mode is the equivalent of diamond_jobs which the CLI already exposes, may be something you want to similarly remove from CLI? Separate issue.

MichaelBarnett:
Adding mt-node to the settings in protein taxonomy blast, and nucleotide taxonomy blast to the yaml settings would be the best way to parse this info to the respective steps. I agree to avoid adding unnecessary cli arguments for this.  That is exactly what the yaml is for, so these low level overrides would be best simply added to the default yaml, and then they can easily be captured in commec where needed.

PR #126, opened 2026-06-29
+allow mt mode to be changed from yml

saarantras:
Following from #124, but manages mt_mode through a yml, and so addresses the cli cruftyness issue. Let me know what you think.

### DIFF
--- a/commec/screen-default-config.yaml
+++ b/commec/screen-default-config.yaml
@@ -25,9 +25,6 @@ databases:
       path: "{default}taxonomy/"
 threads: 1
 diamond_jobs: null
-# Multithreading mode passed to blastn/blastx as -mt_mode (0, 1, or 2; 0 = auto).
-# On fast local disk, 0 performs well across query sizes; for slow networked
-# storage (e.g. HPC/NFS) 1 tends to be faster. Default 1 preserves prior behaviour.
 blast_mt_mode: 1
 do_cleanup: False
 force: False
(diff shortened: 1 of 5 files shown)


commec/screen-default-config.yaml
```diff
@@ -25,6 +25,10 @@ databases:
       path: "{default}taxonomy/"
 threads: 1
 diamond_jobs: null
+# Multithreading mode passed to blastn/blastx as -mt_mode (0, 1, or 2; 0 = auto).
+# On fast local disk, 0 performs well across query sizes; for slow networked
+# storage (e.g. HPC/NFS) 1 tends to be faster. Default 1 preserves prior behaviour.
```
MichaelBarnett: Lets leave comments out of the default yaml for now, as this is what is copied to users for editting.

MichaelBarnett (changes requested):
Lets remove the comments in the default yaml, but also this needs additional error checking.

In commec.config.screen_io::setup() we perform additional checks for some of the user inputs.

Add a check to that function that ensures that mt-mode is set to 0, 1, 2 and if it is out of bounds, an error should be logged, and the program terminated. The user should not be able to get away with adding "A" or "4" to the mt-mode parameter.

The way we update the search handlers seems like the best way to do it, nice.

commec/screen-default-config.yaml
saarantras: Oops. Implemented now.

PR #129, opened 2026-07-08
Commec Setup Update

MichaelBarnett:
## Background

All the databases are updating their home to R2. commec setup required a redesign to more easily allow automation and also download the latest databases in a version controlled manner. This PR addresses that.
Changes

commec setup works autonomously, use -m for a mock-run.

commec setup also can accept a .json file (in the same format as an output commec screen .json) using the cli -r/--restore, to revert or update to those database revisions.

Furthermore, commec screen will now check existing installed databases for possible updates with the config yaml `auto_update_databases: True` setting. Defaults to false.

Includes some minor biorisk and low concern annotation file checking during database validation steps.

The output JSON has been configured to accommodate the revision information of the databases, and likewise, all database-related information has been moved to a posterior location at the end of the JSON. Including all control list and search tool info.

Search tool info has been tidied for erroneous white space characters and special characters.

A Diamond database is still possible, but lives under protein best_match instead of its own path.

## New features

   * New databases are download at the designed location or via yaml.
   * Commec screen can auto update existing databases when directed via yaml.
   * Databases can be downloaded or reverted according to an output or hand crafted json that contains the database_info : revisions information.

## Breaking changes

   * The expected location of databases in the default yaml has changed, requiring updated databases - thankfully, this is fixed in a matter of minutes by simply running commec setup.
   * The JSON output has been significantly changed for database search tool info, revisions, and control list information.


manu-script (changes requested):
The user experience of setting up commec from scratch is now much easier and better with this PR. Kudos @MichaelBarnett!

A useful feature to have in the commec setup is the ability to download a specific version (revision) of the database release. By default, it should fetch the latest revision by reading the latest.json from databases.commec.io, but an optional arg (--revision) could be specified to fetch the corresponding versions of the database components from a previous release from now on. This would allow us to automate benchmarking pipelines to track performance changes across database revisions. It would also likely require us to slightly modify the latest.json to track database components for each release, in addition to the latest revision info.

MichaelBarnett:
Changes made @manu-script  - note also the changes to the JSON output. As Tessa really wanted the control list info out of the way from the top - and so I moved all the database related info down to the bottom of the json in its own folder.

Seems to be working with all of my current tests of all the various ways to try update or not update or revert from jsons or yamls or -d etc etc.

Have a look and squash merge if okay - I've updated the above description so the merge text is reasonable.

manu-script (approved):
Awesome! All the different ways of specifying database paths and revisions seems to be working fine. The experimental and restore options will save us from a lot of headaches in the long run. Thanks for considering and implementing them! :)

PR #130, opened 2026-07-08
Add walk-up kiosk / LAN screening GUI

saarantras:
## Background
Adds a gui so one can run `commec screen` from a walk-up appliance (kiosk) or over the LAN, fully offline. Wraps the CLI; no screening logic changes. Almost entirely additive (new `gui/` dir); the only existing-code change is `commec/utils/template.html`.

I am not sure if this is actually the right way to manage this code. I think splitting GUI off from commec-in-a-box is probably the right move because some users might want to use the GUI without the whole image-minting pipeline thing. Putting this as part of the main common mechanism repo seems reasonable (it is pretty tightly coupled with common-mechanism), but it is almost totally additive, so one could alternatively put it in a different repo. What do folks think? ("We should put this code in a different spot XYZ" would also be a good outcome for this PR).

**Dependencies:** the server runs in the same conda env as commec and needs **flask**, **werkzeug** (via flask), and **pyyaml**; **psutil** is optional (used for system meter testing on osx, and just degrades to "--" if absent instead of breaking). These are *not* added to `environment.yaml`/`pyproject.toml` to avoid bloating the main commec conda env.

### New features
* GUI (`server.py`, `index.html`): paste / upload / USB / server-path inputs, screening presets, live progress, results + linked HTML report.
* Per-device HTTPS (mkcert); LAN clients gated behind a password (localhost walk-up exempt); read-only USB auto-mount.
* **Offline report charts:** vendors `plotly-3.0.1.min.js` and rewrites commec's CDN plotly `<script>` tags to the local copy, so `*_summary.html` renders charts with no internet (license in `vendor/NOTICE.md`).
* Ships deploy docs (`README.md`, `deploy-notes.md`).

### Bug fixes
* `commec/utils/template.html`: HTML-escape (`| h`) query name/description/rationale in the report to prevent HTML injection from user-supplied sequence headers (seems extremely unlikely, but technically possible).

### Breaking changes
* None (I hope & believe).
 
_(n/a)_


MichaelBarnett:
Will need to look through and review, but regarding keeping a GUI as part of the commec package, my instincts tell me this is the *right move*, (but would also like @alexanian  weigh in) to keep them in the same repo. Any major CLI changes to commec would likely require GUI changes at that step, and it would make sense to keep them together. Repo size / download/ / install size is really not a prohibitive issue where we are at currently.

saarantras:
added gui into package proper as @MichaelBarnett suggested. I included the deps in the main package, but if you don't install the flask server it should still work

saarantras:
updated with changes to `develop`.

saarantras:
Other than one issus I just fixed, ruff is complaining about lint issues in files not touched by this branch (just general pre-existing stuff from `devel`.) I could fix them, but it's out of scope for this branch, so I recommend we merge without addressing it here, then open a different branch to do the linting stuff.

MichaelBarnett:
Looks to be in a good place, agree that the Ruff issues are not at play here and are outside scope. Been through several rounds of good feedback and improvement and I really like where we are at.
There is always "more" work to do in refinement, but I think this is in a state of completion enough for 2.1.

commec/gui/vendor/NOTICE.md
MichaelBarnett, line 1: Remove this, plotly no longer used. Update for any other 3rd parties used.

MichaelBarnett: Note this might actually apply to the entire "vendor" file.

commec/gui/vendor/plotly-3.0.1.min.js
MichaelBarnett, line 1: delete?

commec/cli.py
```diff
@@ -26,19 +27,32 @@
...
 )
 from commec.setup import (
     DESCRIPTION as setup_DESCRIPTION,
     add_args as setup_add_args,
     run as setup_run,
 )
 
-from commec import __version__ as COMMEC_VERSION
-
+# The GUI subcommand pulls in Flask (an optional dependency). Import it
```
MichaelBarnett: Whilst this behaviour has been neat up until now, once its in commec conda version, it will no longer be needed. Perhaps just get rid of this check? Flask is tiny afterall.

commec/cli.py
```diff
@@ -86,6 +106,13 @@ def main():
         setup_run(args)
     elif args.command == "list":
         list_run(args)
+    elif args.command == "gui":
+        if gui_run is None:
```
MichaelBarnett: Same here, don't really need the dependency checks.

MichaelBarnett (approved):
Approved with the deletion of some unused parts.

The gui itself runs really well locally for me, and is awesome.

commec/gui/vendor/NOTICE.md
saarantras: Agreed. Though maybe we should put some licensing stuff for the fonts somewhere?

MichaelBarnett (approved):
lgtm, and works well. Will merge it in for 2.1 release this week.

PR #132, opened 2026-07-11
Vectorise the NC coordinate calculations for check_reg_path

MichaelBarnett:
## Background
There was an issue with some entries of a multiquery fasta input being missed as flags, which also brought up a incorrect write warning with copies of dataframes from pandas. This ensures that intead of iterrows we use vectorisation of the non-coding coordinate calculations which is a much better way to handle it and may of addressed the error, subject to a repeat test on another machine.


manu-script (approved):
Can confirm this fixes the issue of warnings.

PR #134, opened 2026-07-17
Update HTML screening report for release

alexanian:
Replaces the Plotly/Mako `<prefix>_summary.html` with an interactive report implemented from the Claude Design prototype.

## It looks pretty slick:
image

- Status tabs — **Flag / Warning / Clear / Skipped** — with counts
- Per-sequence table with severity mini-maps + one-line findings
- Expandable detail: swimlane lanes (Biorisk / Best match / Annotations), a rail of all hits grouped Flag/Warning/Clear, a best-target hero card, and control-list drill-down (jurisdiction / list / listed entry)

## How it works
- **`json_html_output.py`** rewritten: `build_report_model(ScreenResult)` flattens queries → hits → free-form `annotations` into the renderer's data model (taxonomy best-target from `controlled_taxonomy`, biorisk `regulated`/`domain`, low-concern, per-hit control-list join by name, search-tool version parsing). `render_report_html` assembles one self-contained HTML file (embeds the model as JSON + inlines CSS/JS/logo assets). Plotly/pandas/mako removed; `generate_html_from_screen_data` / `_json` / `main` signatures unchanged (so `screen.py` is untouched).
- **`report_assets/`** (new): `report.js` (vanilla renderer ported from the design — string builder + one delegated click handler, no framework), `report.css` (open-licensed **Manrope + Crimson Pro** via Google Fonts, graceful fallback to Arial/Georgia), `commec-logo.svg`.
- **Disposition & tabs derive from commec's authoritative per-hit statuses** — no client-side scope re-classification — so the report always agrees with commec's flag counts. Cleared hits are distinguished from regulated-elsewhere and non-unique matches.
- `pyproject.toml` package-data now ships `utils/report_assets/*`; `template.html` removed.

## Verification
- `pytest commec/tests/test_screen.py` → 5 passed; full suite (290 tests) collects with no import errors; `test_json.py` passes.
- Rendered real v0.6 outputs end-to-end and validated in a headless browser: no console errors; tabs/expand/hit-select/drill-down all work; Flag/Warning/Clear/Skipped counts match commec's own status counts.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


LucasBoldrini:
Looking good! Minor addition that I suggest is adding flags to the left of the control lists so that people have a better sense of what each jurisdiction the control lists refer to before clicking on it. Something like:

🇿🇦 ZAEC Declaration of Certain Biological Goods and Technologies as Controlled Goods

MichaelBarnett:
I can think of some improvements, however I am also inclined to accept as is: 
a) It works, 
b) we can also improve more on minor releases in the near future. 
c) Its superior to the previous html report - especially in how it handles the control lists complexity and that is a key feature of the 1.1 release.

Some things I would like to see in the near future, but perhaps not required for this release, in order of complexity:

* Include a way to "collapse, unfurl" all visible queries.
* Add a "hidden" tab that appears when you hide particular queries, which then banishes them to this tab.
* I instantly preferred viewing at ~133% zoom. Much of the text is too small to be comfortably read with the low contrast between the gray and white used. A font size overhaul is probably a good idea.
* Maybe its coming from the old html visualiser, but the overlap of the hits within a step - especially within annotations - can quickly get overwhelming. It would be nice if when selecting a hit within a query via the scroll bar, that hit visually forced the other hits to no longer collide with it, and also somewhat highlighted the other hits that touched it. This would also be better to see how low concern cleared a particular hit (Do we consider not showing low concern hits as their own hit and simply append their information to any hit that was cleared by them?)

Other than those short ideas, I had a quick squiz at the code and am pleasantly surprised at its brevity, although no javascript expert over here.

Whilst I like the idea of a visual representation of using a Flag for each control lists, there will be the pain point of finding a flag for The Australia Group - although perhaps Aussies flag is fine for that haha. A more fun and perhaps striking solution would be a mini world map, which is highlighted with the regions from each control list hit, and clicking the region highlights the control list that formed it, and vice versa for clicking a control list item.

I approve the PR in current form in the interest of soldiering forward with the 1.1 release.

MichaelBarnett (approved):
Approved as is. See other comment for thoughts.

manu-script:
This HTML update feels like a breath of fresh air, looks very elegant, and provides a lot of useful information! Well done!

I tested this on a 1000-query dataset, and it holds up well. The separate tabs feature makes it so much easier to visualize large query sets. 

Some improvements that could be made, either in this version or in a future update:

a. There seems to be an extra special character at the end of some species names.
image
image
  

b. The width of the bars in the hit overview for biorisk flag/warnings is smaller than the bars of best_match and annotations. I don't know if that's intentional, but it makes biorisk seem less important.
image

c. While the `WARNING` is currently only used for virulence factors, we will also use it for hypothetical proteins in the next release. It may be better to broaden the scope of the warning status description used at the bottom of the page.
image


d. The "Non-unique match" annotation type may not be required for the new best_match db, as they now come under the UniRef100 clusters, which always pass due to non-controlled cluster taxids, and never show up on the hit list of the HTML report.
image

e. Finally, what we always wanted to do, but never got around to: hyperlinking the protein accessions to UniProt and nucleotide accessions to NCBI pages.

alexanian:
Thanks for all of these comments, reviewing and addressing now!

manu-script:
If I may add one more request, it would be nice to also include the revisions of all the database components in the header, which are available in the new JSON format. This would make it easier to troubleshoot bug reports.
image

alexanian:
Alright, I (*mostly Claude) have implemented much of what was requested. 

### Changes

* Indicate list jurisdictions: I thought flag emojis would be a bit too colorful, but I have reinstated country codes, so they chips now look like this:
 
image

* Include a way to "collapse, unfurl" all visible queries
image


* Increased all specified text sizes under 20px by 2px, which I think does improve readability
* Clean special characters before displaying hit names
* The width of the bars in the hit overview for biorisk flag/warnings is now the same as for best match
* Add links to UniProt and NCBI for best match hits
* Added some jiggling of overlapping hits: not totally satisfied with the visuals of this but I think it's useful so let's ship it
 
image

* Updated the `STATUS` descriptions, including a broader definition of warning, and removed "non-unique match" annotation type
image

* Update the header to include database revisions
image

### Things I didn't implement
because I want to think about them a bit more, but I agree they are good ideas that we should ship in the next version of the HTML report:

* Add a "hidden" tab that appears when you hide particular queries, which then banishes them to this tab.
* Do we consider not showing low concern hits as their own hit and simply append their information to any hit that was cleared by them? --> we should discuss, but, yeah, maybe.

PR #135, opened 2026-07-17
Fix comment-character bug in BLAST tabular parsing

manu-script:
## Background
A BLASTN hit whose subject description contained a literal # was silently being misparsed with no error by pandas, coercing the subject taxids in the df into float64, and breaking the controlled taxid lookup logic without any error. This PR squashes the bug by slightly changing the output format of BLAST.

## Changes

- BLASTN/BLASTX ran with -outfmt 7, whose #-prefixed header/comment lines were stripped via pd.read_csv(..., comment="#") — but pandas' comment= truncates a row at the first # found anywhere in the line, not just at line start. A real hit whose subject title legitimately contained a # was silently corrupted or dropped. Switched to -outfmt 6 (no comment lines at all, matching DIAMOND) and dropped comment="#" entirely.
- Moved validate_output() (file-exists check — an empty output file is a legitimate "zero hits" result under tabular-only output) from DiamondHandler up into the shared BlastHandler base class so BLASTN/BLASTX get the same correct handling.
- readhmmer()/readcmscan() had the identical issue ("#" in line check), fixed with startswith("#") for precaution, and removed the now-redundant # Program: ... break since the general comment check already covers that trailer block.
- Updated test fixtures to drop the synthetic outfmt-7-style headers they were injecting.


MichaelBarnett (approved):
Works as intended. Errors on # containing blast outputs from previous runs. Correctly treats empty file as a file with no hits with --resume. And most importantly of all, doesn't get confused by # character containing descriptions.

PR #136, opened 2026-07-19
Update README and remove `commec split`

alexanian:
## Background
Update the README according to all the updates planned for our next release, and remove the `commec split` command, which I think we should exclude given that we actually prefer that people screen multiple queries at once.

### Breaking changes 
* Remove `commec split`


MichaelBarnett (approved):
Simple changes to modify the readme to updated information.
Also removing the commec split cli option. Which is likely a good idea to encourage mixed runs which are more efficient.

PR #140, opened 2026-07-22
Release v2.0.0

MichaelBarnett:
## Background
A large feature rich release for commec, centered around the use of updated commec databases for all steps, as well as screening logic that takes into account control lists from several jurisdictions.

Commec now uses a total database size of ~44 GB (uncompressed), with a download size of ~7 GB.
Commec screen now has a time-to-screen of <1 second per sequence, based on internal benchmarking of over 500 sequences. (This is best achieved by running many queries in a single input fasta)

## Changes:
* Overhaul of the output HTML report.
* New JSON output formats.
* Overhaul of `commec setup` for ease of use, and better automation options.
* Added config.yaml setting to allow commec screen to check for and auto-update to the latest databases.
* Adds the retrieval and use of a new set of custom commec databases, vastly improving screen performance.
* Adds 19 control lists across ~10 jurisdictions, including European Union, and The Australia Group.
* Adds the ingestion of jurisdiction/regional control lists, which alter `commec screen` flag and pass behaviour depending on regional context.
* Adds --region cli argument, to apply jurisdictional context to `commec screen`.
* Blast parameter mt-mode can be configured via config.yaml, useful under some hardware situations.
* Removed no-hit warning, as it is no longer applicable with the use of commec best match databases.
* Removed Diamond support, commec now runs fully with BLAST(x/n), HMMSCAN, and CMSCAN.
* Removed `commec split` cli functionality.
* Adds `commec list` as a cli option, for quickly interacting with Control Lists.
* Updated example data.
* Performance improvements to `commec screen` between screening tool calls.

### Bug fixes
* Improvements and stricter parsing of input config and cli parameters to reduce silent configuration unintended use.
* Fixes an issue with comment characters in blast descriptions sabotaging blast results in some circumstances.
* Fixes an issue with incorrect filtering logic for some low concern steps.
* Fixes an issue with non-coding region calculations affecting other queries 
* Improved handling of Fasta query names and descriptions.

### Breaking changes 
* JSON output format update to 0.6.
* Default config.yaml format updated.
* commec split is removed.
* commec cli arguments for Diamond, and Diamond support removed.
* HTML output format updated.


commec/utils/json_html_output.py
```diff
@@ -1,451 +1,517 @@
...
+    Tidy a display string coming from upstream databases. Some names carry markup
+    (sometimes URL-encoded, e.g. ``%3C/i%3E`` for ``</i>``), stray wrapping quotes,
+    or runs of whitespace. Decode, strip tags, and trim so it presents cleanly.
     """
+    v = _clean(v)
+    if not isinstance(v, str):
+        return v
+    v = unquote(v)  # %3C/i%3E -> </i>
+    v = _HTML_TAG_RE.sub("", v)  # drop <i>...</i> and similar markup
+    v = v.replace(" ", " ")  # non-breaking space -> space
```
manu-script, line 79:
```suggestion
    v = v.replace(" ", " ")  # non-breaking space -> space
    v = v.replace("\ufffd", "")  # drop mis-encoded replacement chars from source data
```
Strip � characters from some malformed taxa names in control list.
Image

commec/screeners/check_reg_path.py
```diff
@@ -1,380 +1,567 @@
...
+            len(compliances),
+            len(conditional_compliances),
+            len(warn_compliances),
+        )
+
+        if len(compliances) > 0:
+            regulated_annotations.extend(
+                [_create_hit_info(row, compliances) for i, row in data.iterrows()]
+            )
             continue
```
manu-script, line 344: The three accumulator lists (`compliances`, `warn_compliances`, `conditional_compliances`) are created once, at lines 299–301, before the loop.  Since the loop only ever `.extend()`s them and never clears them per iteration, they carry state from every taxid processed so far. 
Potential issue: The per-taxid row classification inside the loop (for instance at line 341: `if len(compliances) > 0:`) uses the same already-accumulated list from a previous taxid to decide where the current taxid's own rows go, and `_create_hit_info(row, compliances)` at line 343 attaches compliances' metadata to the row. Verify if this logic is correct.

commec/screeners/check_reg_path.py
```diff
@@ -1,380 +1,567 @@
...
+        len(conditional_compliances) >= N_NON_REGIONAL_HITS_TO_WARN
+    )
+
+    if is_controlled:
+        non_regulated_annotations = (
+            non_regulated_annotations
+            + warn_regulated_annotations
+            + regionally_regulated_annotations
+        )
+        return _create_hit_result_from_annotations(
```
manu-script, line 378: Check whether concatenating `non_regulated_annotations` with `warn_regulated_annotations` and `regionally_regulated_annotations` here can wrongly cause a cluster status downgrade from FLAG -> PASS in some conditions because inside `_create_hit_result_from_annotations` (line 478), we have:

```
if len(non_regulated_annotations) > 0:
    status = ScreenStatus.PASS
    ...
    hit_description = f"Mix of {len(regulated_annotations)} ... and {len(non_regulated_annotations)} non-controlled ..."
```

manu-script (changes requested):
The release branch has been pretty stable in my test runs. Some minor changes in `commec/utils/json_html_output.py` and precautionary checks in `commec/screeners/check_reg_path.py` are suggested below. Otherwise, it looks ready to be released!

AgentK9:
Stupid merge conflict, my bad

commec/screeners/check_reg_path.py
MichaelBarnett: This is correct logic. The code is collecting and parsing all control list information from a region, storing the overall status of the region by extending the following lists:
```
    regulated_annotations = []
    regionally_regulated_annotations = []
    warn_regulated_annotations = []
    non_regulated_annotations = []
```
These lists are then interrogated and used after lines 372.

MichaelBarnett: Ah yes, the specific checks on the compliances etc lists do need to occur on a in-loop local var before appending to the other lists. This would be causing a waterfall effect that continues to mark entitites as flagged as soon as one is.

MichaelBarnett: fixed.

commec/screeners/check_reg_path.py
MichaelBarnett: This is a design choice, such that regionally exempt outputs, can contribute to a mixed-non-regulation result.
We trying to ensure that greatest priority wins, for example, If a Flag result, is cleared with mixed results. The hit of "Mixed Flag" should take priority over giving the user a Pass of 2 regionally exempt hits.

In some ways this is a hangover from when we were considering passing a Warn when 2 regionally exempt hits or more showed up. But we went in favour of Pass, and still create the hit to report on instead - similar to mixed.

The warn logic is actually not used - we don't have any control lists in a Warn capacity although anyone could author one and it would run, (future us will be using this for the hypothetical proteins). If Warns were present for the current code, It is possible for a Flag to be downgraded to a Pass through the addition of a compliance warn hit, which maybe should be downgraded to a Warn instead...

I will remove specifically the adding the warn hits (which don't exist yet anyway), as at the very least they wont be used to "clear" a Flag as Mixed, and it might serve as an anti-gotcha in the future when we implement the hypotheticals. Although this just creates the reverse gotcha, what really needs to happen is that it needs to be communicated via some sort of extra flag passed into _create_hit_result_from_annotations ... something I was trying to avoid for simplicity in designs sake. I'll fix.

I believe we definitely still want the regionally exempt hits to remain as able to clear due to "mixed" at this stage.

commec/utils/json_html_output.py
MichaelBarnett: This is more an issue with including that bad character in the control_list, I will update there instead. I would rather catch and display these rather than filter and hide them. The Display_Name field in control lists should be clear of miscoded chars.

PR #142, opened 2026-07-28
Build Custom Container & conda recipe fixes

AgentK9:
## Background

Since bioconda takes a while to merge, we wanted to create our own container. This MR creates that container in CI, and also fixes some issues found by claude in the conda recipe.

**Issues**:

### Bug fixes
* Fix conda recipe drift

### New features
* Adds commec container built in CI - now only requires a tag on the common-mechanism repo, rather than merge in bioconda recipes repo.


alexanian:
Would you be happy to split the ruff run from the container + conda fixes? Just to make it easier to review :)

.github/workflows/container.yml
```diff
@@ -0,0 +1,123 @@
+name: Container
+
+# Builds the commec container image, runs the test suite and a binary smoke test
+# inside it, and publishes to GitHub Container Registry.
+#
+# Pull requests build and test only — no registry login, no push, so PRs from
+# forks (which have no packages:write token) work unchanged. Images are pushed
+# only for pushes to main and for v* tags.
```
MichaelBarnett, line 8: I think we only care about generating a container for version tags? Less so pushes to main - although they should be correlated and the same. I am skeptical that this will always be true, and we don't want to automate a container creation for some minor readme update we want in main?

Or perhaps I am misunderstanding the intent - and we're generating the v_ tag automatically?

MichaelBarnett (approved):
Looks good - super handy to have our own docker image container - I still need to check it all works locally but looks okay.

Just need to confirm that we perhaps only want to make containers form tags only, not merges to main?

.github/workflows/container.yml
AgentK9: Sorta depends on what we want `latest` to point at - the latest release or the latest push to main?

Either way, not a huge change - whichever we prefer.

alexanian: I would say latest release (either is probably fine, but latest release should be the version that we've tested and are sure works)

manu-script:
Referring to https://github.com/ibbis-bio/common-mechanism/issues/147#issuecomment-5257942559, let's not merge this until we test the release with bioconda-utils.

PR #144, opened 2026-07-28
Pin ruff

AgentK9:
## Background

Ruff's version needs to be pinned, and checks must be able to fail so that it gates PRs

### Bug fixes
* CI: ruff version is now pinned
* CI: ruff can now fail


alexanian (approved):
This LGTM, thanks for splitting out the import sorting!

alexanian:
Actually, that said, having just had ruff fail for me on https://github.com/ibbis-bio/common-mechanism/pull/145 ― should we run `ruff` as a pre-commit hook?

.github/workflows/pr-checks.yml
```diff
@@ -15,11 +15,16 @@ jobs:
         uses: actions/checkout@v4
       - name: Ruff check
         uses: astral-sh/ruff-action@v4.1.0
+        with:
+          version: ">=0.15.10,<0.16.0"
+      # These two steps must not auto-fix: mutating files in CI and exiting 0
+      # means unsorted imports and unformatted code pass unnoticed. Run
+      # `ruff check --select I --fix . && ruff format .` locally to fix failures.
       - name: Ruff isort
```
alexanian: If I'm understanding this right, what this does is change it so that the CI _fails_ when ruff fails? This comment feels like it may be hard to understand without the context of knowing about this bug, would suggest instead


```suggestion
      # Check ruff run, but do not fix
      # (format fixes in CI will exit 0 and then be lost to the codebase)
      name: Ruff isort
```

.github/workflows/pr-checks.yml
```diff
@@ -15,11 +15,16 @@ jobs:
...
-        run: ruff check --select I --fix
+        run: ruff check --select I .
```
alexanian: Actually, having chatted with Claude, I think we shouldn't add this without adding something about `isort` to our docs or pyproject.

I think we need to:

1. Do the same version pinning in environment.yaml under dev dependencies https://github.com/ibbis-bio/common-mechanism/blob/4ebd5a24a49c0d98ab133e9d9022de3d5d2e11e8/environment.yaml#L22
2. Add the following to pyproject.toml:

```
# In addition to default ruff config, sort imports
[tool.ruff.lint]
extend-select = ["I"]
```

I'm not sure if there's any other non-default configuration we want to add?

alexanian (changes requested):
I did a bit of a longer review, and now have a few changes to request, to hopefully keep all of us sync'd up. Some suggestions here: https://github.com/ibbis-bio/common-mechanism/commit/0b0c84320740f28fe9244f7927aada03f718c8a7

In addition to my comments below, I think we should either:

1. Add documentation to the README that we are using `ruff` for code formatting, telling people to use it.
2. (my preferred solution) add their pre-commit hook: https://github.com/astral-sh/ruff-pre-commit (with the same version pinning we do here, probably?)

```
repos:
- repo: https://github.com/astral-sh/ruff-pre-commit
  # Ruff version.
  rev: v0.16.0
  hooks:
    # Run the linter.
    - id: ruff-check
    # Run the formatter.
    - id: ruff-format
```

This also proposes pinning a minor version because it looks like ruff introduces breaking changes with minor versions: https://docs.astral.sh/ruff/versioning/

AgentK9:
> Actually, that said, having just had ruff fail for me on #145 ― should we run `ruff` as a pre-commit hook?

pre-commit hooks are good but we can't guarantee that anyone will run them - having it in CI requires it no matter if people have pre-commit installed or not.

We can add pre-commit hooks, just seemed out of scope.

.github/workflows/pr-checks.yml
AgentK9: I wish we could use a single source of truth for the ruff version, but yeah these make sense.

alexanian:
yeah, happy to decide that the pre-commit hook is out of scope ― agreed that adding it does not get rid of the need for a CI check, just provides a preferred / low-effort way to pass that CI check by default.

AgentK9:
I think this is all good to go now

.pre-commit-config.yaml
```diff
@@ -0,0 +1,9 @@
+repos:
+- repo: https://github.com/astral-sh/ruff-pre-commit
+  # Ruff version.
+  rev: v0.16.0
```
alexanian: Should we sync this with environment.yaml and pr-checks.yml? But otherwise LGTM

alexanian (approved):
LGTM, though can we also add [the README change proposed here](https://github.com/ibbis-bio/common-mechanism/commit/0b0c84320740f28fe9244f7927aada03f718c8a7#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5)?

Thanks!

alexanian (approved):

PR #145, opened 2026-07-29
Tweaks from testing (output prefixes, commec list)

alexanian:
## Background
A few small fixes found while I was testing 2.0 and updating [our docs](https://github.com/ibbis-bio/common-mechanism/wiki).

### Bug fixes
* Output-prefix paths: `commec screen input.fasta -o /some/other/dir/prefix` was writing to the input file instead of the `-o` location

### Breaking changes 
* `commec list` now accepts `regions` from a YAML file and expects regions to comma-separated, matching `commec screen`


MichaelBarnett (approved):
Runs as intended on my machine, and addresses some important fixes to commec list cli in code.

alexanian:
I'm going to merge this because the linting issues should get addressed in https://github.com/ibbis-bio/common-mechanism/pull/144

PR #146, opened 2026-07-30
Automate new PR's to target 'develop', not 'main'

MichaelBarnett:
## Background
It is super annoying to have to manually remember to specify that a feature branch should target develop, and quite easy to forget in the heat of feature release to ensure features are squashed to develop, not main.

This little script - fully generated by Claude - should address that.


AgentK9:
I think we can just configure the commec repo to disable PRs directly into main except for develop

AgentK9:
Looks like the way to do this would be via "restrictions" in rulesets, which is a GitHub enterprise feature

.github/workflows/retarget-pr-to-develop.yml
```diff
@@ -0,0 +1,25 @@
...
+on:
+  pull_request:
+    types: [opened]
+
+permissions:
+  pull-requests: write
+
+jobs:
+  retarget:
+    if: ${{ !startsWith(github.event.pull_request.head.ref, 'release_v') }}
```
alexanian, line 12: this looks great! we should maybe document somewhere that releases always start with that prefix. thoughts on adding a `/.github/CONTRIBUTING.md` file (based off our internal "commec branching and package release process" doc)? I've attached a draft.

[CONTRIBUTING.md](https://github.com/user-attachments/files/30562241/CONTRIBUTING.md)

MichaelBarnett:
> Looks like the way to do this would be via "restrictions" in rulesets, which is a GitHub enterprise feature

I also considered whether we should just make develop our default branch (Which would auto-direct any new PR' to it) but of course then it would double as the landing page, as well as the default clone branch, which is less than ideal.

Regarding restrictions, I did want to force squash merge into develop, and merge from develop to main, as per standard git flow protocols, which may be well suited to be used via rulesets, but also may be a enterprise feature - albeit I haven't looked into it yet.

.github/workflows/retarget-pr-to-develop.yml
AgentK9, line 12: just so that this doesn't run on every push, even when the PR already targets develop, can we move the `if:` block below up here?

AgentK9 (approved):

.github/workflows/retarget-pr-to-develop.yml
MichaelBarnett: Sure thing, I was under the impression that the 

```
on:
   pull_request:
      types: [opened]
```
Was going to restrict to only running on a new PR, and not every push, but perhaps I am mistaken on that workflow, nethertheless, it should probably fail earlier with the already pointing check and not fire up the whole script.

PR #149, opened 2026-08-07
Fix path creation in commec flag

AgentK9:
## Background
A user reported the following error:

> commec flag needs a pre-existing directory. Errored on a missing output path; the core screen step was unaffected.

This PR aims to fix that.

### Bug fixes
* Create the output directory and its parents up front, rather than failing once all the screen files have been parsed.

### DIFF
--- a/commec/flag.py
+++ b/commec/flag.py
@@ -9,7 +9,7 @@ positional arguments:
 options:
   -r, --recursive      show this help message and exit
   -o, --output OUTPUT_DIR
-                       path where output CSVs should be written
+                       path where output CSVs should be written, created if it does not exist
 
 The screen_pipline_status covers the status of the 4 screening steps, while flags.csv has just
 flag outcomes for various things that commec can flag, and flags_recommended just has an overall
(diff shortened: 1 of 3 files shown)


manu-script (changes requested):
While these are welcome changes to `commec screen -o` behaviour, the original bug report from the user was:
> commec flag needs a pre-existing directory. Errored on a missing output path; the core screen step was unaffected.

This concerns `commec flag -o`, which still throws `commec flag: error: argument -o/--output: ... is not a valid directory path` error when the given output directory doesn't exist.

uv.lock
```diff
@@ -0,0 +1,8 @@
+version = 1
```
alexanian: Do we want to commit this?

commec/flag.py
```diff
@@ -235,9 +235,18 @@ def run(args: argparse.Namespace):
...
     search_dir = args.directory
     search_recursive = args.recursive
-    output_dir = args.output or os.path.dirname(search_dir)
+    output_dir = expand_and_normalize(args.output or os.path.dirname(search_dir))
     evalportal_format = args.evalportal_format
 
+    # The output directory is written to at the end of the run, so create it (and any missing
+    # parents) up front, rather than failing once all the screen files have been parsed.
+    try:
+        os.makedirs(output_dir, exist_ok=True)
```
alexanian:
```suggestion
    try:
        os.makedirs(output_dir, exist_ok=True)
```

classic [zombie hedge](https://x.com/julianboolean_/status/2085443173539819526) comment, not needed now the bug is fixed imo

alexanian (approved):

uv.lock
AgentK9: oops

commec/flag.py
AgentK9: huh interesting term

alexanian (approved):

manu-script (approved):

PR #151, opened 2026-08-12
Removed conda recipe as the actual recipe lives in bioconda fork

manu-script:
## Background
While going through the release process written up by Michael, I realized the conda recipe in the main repo is not used anywhere, and the actual recipe that's used to build our conda package lives in the Bioconda fork https://github.com/ibbis-bio/bioconda-recipes/blob/master/recipes/commec/meta.yaml. As this led to some confusion during the v2.0 release, I propose removing the recipe from the main repo and using only the environment.yaml to track all the dependencies and updating the main recipe in Bioconda during new releases.

## Changes
 - Removed the unused conda-recipe/meta.yaml


MichaelBarnett (approved):
Initial added as a way to prepare major changes to the meta.yaml recipe so they could easily be copied and not forgotten at release when making the bioconda-recipes fork. This plays less of a role and is ultimately redundant.

Removing may reduce confusion of which set of redundancies is the ground truth.

PR #152, opened 2026-08-14
Fix browser tab OOM on large spreadsheet screens

saarantras:
Fixes the OOM issue by trimming down the terminal output display. Also only shows 500 items at a time & adds a "show more" button in case the output list is too long (not strictly necessary, but i want to ease the burden on the RAM and it starts to chug with really long lists. I figure we let the user "opt in" to eating a bunch of ram by clicking on the full report).


commec/gui/index.html
```diff
@@ -1243,13 +1243,84 @@
...
+    logEl.textContent = logHeader() + logLines.join("\n") + "\n";
+  } else {
+    logEl.appendChild(document.createTextNode(chunk));
+  }
   if (atBottom) logEl.scrollTop = logEl.scrollHeight;
 }
 
+/* Drawn in bounded chunks: one innerHTML write for many rows is enough on
+   its own to exhaust the renderer. Order matches the source spreadsheet. */
+const QUERY_CHUNK = 500;
```
alexanian: Does it make sense to try to cluster all the CONSTs in one place? (I actually don't know what the HTML norms are on this)

alexanian (approved):
Thanks for the quick fix!

commec/gui/index.html
saarantras: agreed, changing now

PR #155, opened 2026-08-14
Substitute non-IUPAC characters with 'N' when parsing input FASTA

manu-script:
## Background

Input FASTAs are not always clean. They arrive from spreadsheets, word processors, and web forms, and carry things that are not bases: non-breaking spaces, en-dashes, byte order marks, gap characters from an alignment, inosine, or a protein sequence submitted by mistake. Previously, these were passed through to BLAST and HMMER largely as-is, or substituted in ways that silently changed sequence length.

Two properties matter for screening correctness:

1. **Length must be preserved.** Hit coordinates from the search tools index into the query sequence, so deleting a character shifts every downstream coordinate. Anything unrecognised has to be *masked*, not dropped.
2. **The operator must be warned for a high number of substitutions.** A query that is >10% substituted `N` may still return a clean screen, and that result may be influenced by the substitutions.

This PR normalises queries at parse time and adds warnings when query sequences are substituted.

## Changes

* **`substitute_non_iupac`** upper-cases each query and masks every character that is not an IUPAC nucleotide code with `N`, returning the number of substitutions. Masking rather than deleting keeps hit coordinates valid.

* **Two warnings.** One whenever anything is substituted, and a second when the proportion masked exceeds the new `NON_IUPAC_SUBSTITUTION_THRESHOLD` (0.1), which suggests the input is not a nucleotide sequence at all:

  ```
  Query haemagglutinin_fragment: substituted 13 non-IUPAC nucleotide characters with 'N'
  Query haemagglutinin_fragment: 21.7% of characters were not IUPAC nucleotide codes. This may
  not be a nucleotide sequence; screening it may not give correct results.
  ```

* **Uracil is treated as a valid base**, so RNA queries pass through untouched. The valid set is built from `ambiguous_dna_letters + ambiguous_rna_letters`. blastn maps `U` to `T` natively, and Biopython handles `U` in `translate` and `reverse_complement`, so no transcription step is needed. Masking `U` would have destroyed a significant fraction of any RNA query.


commec/config/screen_io.py
```diff
@@ -262,28 +284,40 @@ def output_yaml(self, output_filepath: str | os.PathLike):
...
+
+        Sequence lines have all whitespace removed, rather than substituted,
+        since it is a formatting artefact rather than a base. Any other non-ASCII
+        character is replaced with an underscore, so that hit coordinates remain
+        valid; Biopython sequences are ASCII-only, so the character cannot be
+        left in place for `substitute_non_iupac` to handle, but an underscore is
+        not an IUPAC code, so it is still masked with 'N' and counted there.
+
+        The input is read as `utf-8-sig` so that a leading byte order mark is
+        consumed by the decoder, rather than being mistaken for sequence.
```
alexanian:
```suggestion
        Write a FASTA, cleaning headers of special characters and sequences of
        whitespace, non-ASCII characters, and non-IUPAC codes.
        
        Headers have non-ASCII whitespace (such as non-breaking spaces) and `#`
        characters replaced with underscores, since Biopython finds record id by
        splitting headers on Unicode whitespace.

        Sequence lines have all whitespace removed. Any other non-ASCII character
        is replaced with an underscore, so that hit coordinates remain valid. The
        underscores then are replaced with 'N' by `substitute_non_iupac` masking.

        The input is read as `utf-8-sig` so that a leading byte order mark is
        consumed by the decoder, rather than being mistaken for sequence.
```

alexanian: I accept the utf-8-sig is reasonable here, since AIUI if there isn't a byte order mark, it's just read as utf-8, but I don't have a lot of familiarity with that encoding format.

alexanian (approved):
One suggestion to shorten a docstring but generally LGTM

PR #156, opened 2026-08-15
Update env name

AgentK9:
## Background

Rename the environment to be deployed to to be less alarming


.github/workflows/regression.yml
```diff
@@ -22,7 +22,7 @@ concurrency:
 jobs:
   dispatch:
     runs-on: ubuntu-latest
-    environment: infohazard        # <-- the required-reviewers gate lives here
+    environment: IBBIS-Staff-CI        # <-- the required-reviewers gate lives here
```
alexanian, line 25: Thank you for the quick fix, Deployments "Infohazard" was stressing me out haha 

Image

alexanian (approved):

PR #157, opened 2026-08-18
compatability with CIAB

saarantras:


commec/setup.py
```diff
@@ -416,7 +421,11 @@ class CommecDatabaseUpdater:
...
 
-    def __init__(self, existing_location: os.PathLike, base_url: str):
+    def __init__(
+        self,
+        existing_location: os.PathLike,
+        base_url: str = R2_PUBLIC_BASE_URL,
+    ):
         self.name = None
         self.write_location = existing_location
         self.base_url = base_url
```
manu-script: This default binds at import, so it can't track `screen-default-config.yaml` or be patched. Resolving in the body fixes both, and bare `CommecDatabaseUpdater(path)` calls — which `commec-db-status.py` in CIAB relies on — keep working unchanged.

```suggestion
    def __init__(self, existing_location: os.PathLike, base_url: str | None = None):
        self.name = None
        self.write_location = existing_location
        self.base_url = base_url or YamlIO.default_base_url()
```

commec/setup.py
```diff
@@ -379,7 +382,9 @@ def fetch_revisions_from_json(filename: str | os.PathLike) -> dict | None:
     return db_revisions
 
 
-def fetch_latest_revisions(base_url: str) -> dict | None:
+def fetch_latest_revisions(
+    base_url: str = R2_PUBLIC_BASE_URL,
+) -> dict | None:
```
manu-script: Same fix needed here, but it spans two places and only one is in this diff — `raw = fetch_r2_object("latest.json", base_url=base_url)` on line 393 is unchanged context, so this can't be a one-click suggestion. Applying a `None` default without the body change would pass `None` straight through to `fetch_r2_object`, so both parts have to land in the same commit:

```python
def fetch_latest_revisions(base_url: str | None = None) -> dict | None:
    """
    A latest.json manifest exists at the root of the R2 bucket, listing
    the latest revision of each database. Downloads and parses it into a
    dict. Returns None if it could not be fetched or parsed.

    base_url defaults to the packaged config's host, resolved per call so it
    tracks that file rather than whatever it said at import.
    """
    raw = fetch_r2_object("latest.json", base_url=base_url or YamlIO.default_base_url())
```

manu-script (commented):
Resolve the default base URL lazily instead of binding it at import.

manu-script (changes requested):

commec/setup.py
saarantras: good point. Tossed up a new commit, it should patch this.

manu-script (approved):
This PR restores the three symbols commec-in-a-box imports (`R2_PUBLIC_BASE_URL`, `fetch_latest_revisions()`, `CommecDatabaseUpdater(path)`) as optional-arg defaults resolved lazily from the packaged config, and extends the GUI presets fallback to the appliance's `WorkingDirectory` layout. All four branches covered by new tests.

PR #161, opened 2026-08-19
Retarget only main pull requests

saarantras:
Only pull requests targeting main are automatically retargeted to develop.

PR #153, opened 2026-08-14
Control list output hotfix

MichaelBarnett:
## Background
Updates the column headings of the output csv for commec list to be a) inline with the blog post we're writing, and b) easier for a human to understand:

CN_DUPRC,ZA_EC,AG_CCL,AG_PPCCL,IN_SCOMET,EU_LAP,EU_DUECR,RU_RFECHA,RU_RFECA,US_BSATHHS,US_BSATUSDA,US_PPQUSDA,US_CCL,US_ML,BR_CIBES,UK_SAPO,UK_ATCSA,UK_COSHH,UK_SECL

In short, the region acronym is appended to the front, AND is removed from the front of the list acronym if it was there. i.e. USCCL becomes US_CCL instead of US_USCCL. etc.

PR #163, opened 2026-08-20
Release v2.1.0

manu-script:
# Release v2.1.0

## Background

Release PR for `commec` v2.1.0, merging the accumulated work on `develop` into `main`. This is the first release since [v2.0.0](https://github.com/ibbis-bio/common-mechanism/releases/tag/v2.0.0) (4ebd5a2).

The headline addition is a **walk-up kiosk / LAN screening GUI** (`commec gui`), which lets an operator run screens from a browser on an offline appliance without touching the CLI. Alongside it: a configurable database download host for `commec setup`, input FASTA normalisation for non-IUPAC characters, a Google Colab demo notebook, and a set of bug fixes found while testing 2.0 and writing the user docs.

**Version**: `2.0.0` → `2.1.0` (`pyproject.toml`)

### New features

* **Screening GUI (`commec gui`)** (#130) — a browser front-end for running `commec screen` from a walk-up appliance (kiosk) or over the LAN, fully offline. Wraps the CLI; no screening logic changes.
* **Configurable database host for `commec setup`** (#154) — `commec setup --base-url <url>` and a `base_url` config key override where databases are downloaded from. Better download error handling and TLS failure notes with the `openssl s_client` check and `SSL_CERT_FILE` override.
* **Non-IUPAC character substitution on FASTA parse** (#155) — input queries are upper-cased and every non-IUPAC nucleotide character is *masked* with `N` (masked, not deleted, so hit coordinates stay valid). 
* **Google Colab demo notebook** (#148) — [notebooks/commec_colab_screen.ipynb](notebooks/commec_colab_screen.ipynb) runs `commec setup` + `commec screen` from a simple input form, linked from the README so anyone can demo commec without a local install.

### Breaking changes

* **`commec setup -y/--config` is removed** (#154). `-d/--databases` is now required and is the only way to say where databases go. This prevents users passing the generated example config back into setup, which would overwrite it.
* **Setup now writes `example-config.yaml`, not `config.yaml`** (#154), and **overwrites it on every run** rather than writing it only when absent. It is a file to *copy and edit*, then pass to `commec screen -y` — nothing reads it automatically. 
* **`commec list -r/--regions` now takes a comma-separated string**, not a space-separated list (#145) — e.g. `-r 'NZ,EU'` — matching `commec screen`. Regions can also be supplied from a YAML file.
* **`commec list` CSV column headings renamed** (#153) — the region acronym is prefixed and de-duplicated from the list acronym (`USCCL` → `US_CCL`). 

### Bug fixes

* **Off-by-one in the minimum query length gate** (#162) — a query of exactly 41 bp was translated and protein-screened, but never written to `nt_path`, so nucleotide search was being silently skipped.
* **Output-prefix paths ignored** (#145) — `commec screen input.fasta -o /some/other/dir/prefix` was writing next to the input file instead of the `-o` location.
* **`commec flag` required a pre-existing output directory** (#149) — the output directory and its parents are now created up front, rather than failing after all screen files have been parsed.
* **Downloaded archives left behind** (#154) — the `os.remove` of `<db>.tar.zst` after extraction had been left commented out, so every update left a full-size archive on disk.
* **Browser tab OOM on large spreadsheet screens** (#152) — GUI terminal output is trimmed and long output lists paginate at 500 items behind a "show more" button.
* **HTML injection in the report** (#130) — query name, description, and rationale are now HTML-escaped (`| h`) in `commec/utils/template.html`, so a user-supplied sequence header can't inject markup.

### Refactoring

* **Removed the unused `conda-recipe/meta.yaml`** (#151) — the recipe that actually builds our conda package lives in the [Bioconda fork](https://github.com/ibbis-bio/bioconda-recipes/blob/master/recipes/commec/meta.yaml). Keeping a second copy here caused confusion during the v2.0 release; `environment.yaml` is now the single in-repo dependency record (#159).
* **Ruff formatting and import sorting** (#144, #150) — ruff is pinned (`>=0.15.10,<0.16.0`) and CI checks can now fail and gate PRs.
* **PR auto-retargeting** (#146, #161) — new pull requests that target `main` are automatically retargeted to `develop`; PRs already aimed elsewhere are left alone.
* **commec-in-a-box compatibility** (#157)
*  **deployment environment rename** (#156).

## ToDo

- Update README & Wiki to reflect the new GUI interface @saarantras 
- Full functional regression testing @AgentK9


saarantras (approved):
LGTM. Some nitpicks, but grist for other PRs, should not block release.

saarantras (approved):

PR #164, opened 2026-09-06
Add amino acid (protein) sequence input support

krishnanj:
## Summary

Adds amino acid (protein) sequence input support to COMMEC. Users can submit a protein FASTA via \`commec screen --protein\` or via a new GUI checkbox. The pipeline routes protein sequences directly to hmmscan and blastp, skipping 6-frame translation and all nucleotide-only steps.

Addresses Manu's PR #155, which rejects amino acid input. This PR reverses that rejection by implementing a proper protein code path.

## What changed

**Pipeline routing**
- \`--protein\` CLI flag declares all input sequences as amino acids. Sequences are written to a protein FASTA and passed to hmmscan and blastp directly.
- Molecule type is determined solely by \`--protein\`; there is no auto-detection.

**Steps skipped for protein input**
- Nucleotide taxonomy (Step 3): \`screen_nucleotides()\` exits early and marks the step SKIP. Without this guard, \`calculate_noncoding_regions_per_query\` sees no blastx hits, treats the entire sequence as non-coding, and passes protein sequences to blastn, which is incorrect.
- Nucleotide low-concern (blastn and cmscan, Step 4): no nucleotide FASTA exists for protein input; steps are marked SKIP.

**HTML report**
- Hit coordinate labels now show "aa" for protein queries and "bp" for nucleotide queries.

## Files changed

| File | Change |
|------|--------|
| \`commec/screen-default-config.yaml\` | Add \`protein_input: False\` default |
| \`commec/screen.py\` | Add \`--protein\` arg; skip nucleotide taxonomy for protein input; set \`qr.is_protein\`; use \`MINIMUM_PROTEIN_QUERY_LENGTH\` for AA queries |
| \`commec/config/screen_io.py\` | Route protein sequences via \`--protein\` flag only; remove auto-detection |
| \`commec/config/constants.py\` | Add \`MINIMUM_PROTEIN_QUERY_LENGTH\` (= \`MINIMUM_QUERY_LENGTH // 3 = 13\`) and \`BIORISK_SHORT_QUERY_AA_THRESHOLD\` (= \`BIORISK_SHORT_QUERY_NT_THRESHOLD // 3 = 66\`) |
| \`commec/config/result.py\` | Add \`is_protein: bool = False\` to \`QueryResult\` |
| \`commec/screeners/check_low_concern.py\` | Accept \`None\` for \`rna_handler\` and \`dna_handler\` |
| \`commec/screeners/check_biorisk.py\` | Apply per-row E-value threshold in \`biorisk_evalue_filter\` based on \`frame\` column (frame 0 = protein, uses AA threshold) |
| \`commec/utils/json_html_output.py\` | Pass \`is_protein\` to JSON hit model |
| \`commec/utils/report_assets/report.js\` | Display "aa" vs "bp" from \`isProtein\` |
| \`commec/tools/blastp.py\` | New \`BlastPHandler\` (mirrors \`BlastXHandler\`) |
| \`commec/tools/hmmer.py\` | Fix KeyError for protein query names ending in \`_N\` (e.g. \`ORF_1\`); set \`frame=0\` for protein-input rows |
| \`commec/gui/index.html\` | Add "Sequence type" checkbox |
| \`commec/gui/server.py\` | Inject \`protein_input: true\` into per-run YAML config |
| \`commec/tests/test_aa_input.py\` | Five end-to-end factory tests for protein pipeline |
| \`commec/tests/test_io_params.py\` | Add \`protein_input: False\` to fixture dicts |
| \`commec/tests/test_screen_io.py\` | Remove auto-detection test; update substitution-threshold test to reflect correct behaviour without auto-detection |
| \`commec/tests/test_check_biorisk.py\` | Add \`frame\` column to all test DataFrames; add tests for protein short-query E-value path |
| \`commec/tests/test_gui_inputs.py\` | Add tests for protein_input propagation and absent-when-unchecked |

## Test results

All 361 tests pass. \`ruff check --select I --fix\`, \`ruff format\`, and \`pytest -vv\` all pass on commit 993902f.

**Query file**

\`data/protein_test_seqs.fasta\`: 10 protein sequences from the Virulence Factor Gene Database (VFDB), covering a range of virulence factor types and source organisms (Acinetobacter baumannii, Bacillus anthracis, Bacillus cereus, Clostridium perfringens, Legionella pneumophila, Chlamydia trachomatis, and others). Total length: 4,747 amino acids.

**Reference databases (best_match v1.0)**

| Database | Tool | Contents | Size |
|----------|------|----------|------|
| \`best_match/protein/\` | blastp | COMMEC Protein BLAST Database: 25,524,824 sequences, 10.1 billion residues | 15 GB |
| \`best_match/nucleotide/\` | blastn | COMMEC Nucleotide BLAST Database: 3,000,821 sequences, 86.7 billion bases | 21 GB |

The protein database is derived from UniRef protein clusters. The nucleotide database covers viral reference genomes. Both are used to identify the closest matching organism for each query region, which is then checked against the jurisdiction control lists.

**End-to-end screen, biorisk only (\`--skip-tx --protein\`), 10 VFG sequences:**
- 6 Pass, 3 Warning (virulence factor), 1 Flag (regulated gene)

**End-to-end screen, full taxonomy (\`--protein\`, best_match v1.0 installed):**

| Result | Count | Sequences and reason |
|--------|-------|----------------------|
| Flag | 3 | VFG016254 (regulated gene, biorisk); VFG002278 (Clostridium perfringens TaxID 195103, controlled; also virulence factor); VFG045607 (Legionella pneumophila TaxID 272624, controlled) |
| Warning | 2 | VFG037837, VFG007946 (virulence factor, non-regulated organism) |
| Pass | 5 | VFG000682, VFG051435, VFG045335, VFG001380, VFG000200 |

Legionella pneumophila and Clostridium perfringens were correctly identified as regulated organisms via blastp taxonomy. Nucleotide taxonomy was skipped cleanly (marked SKIP) for all 10 sequences. The biorisk step correctly detected virulence factor signatures in sequences from VFDB, which provides informal ground truth that the flags are biologically meaningful.

A sequence named \`ORF_1\` was tested manually and processes without error; the KeyError fix in \`hmmer.py\` works correctly.

## Notes

- Mixed FASTA input (protein and nucleotide sequences in the same file) is not supported; a single molecule type per file is required by FASTA convention. The test for this was removed.
- \`commec setup\` installs the best_match database (~36 GB total) required for full taxonomy screening.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


commec/screen.py
```diff
@@ -779,10 +841,24 @@ def screen_low_concern(self):
...
+        # blastn and cmscan require a nucleotide FASTA. For protein-only input,
+        # protein records are not written to the nucleotide FASTA, so those files
+        # are empty. Skip both steps and pass None to the parser.
+        protein_only = self.params.config.get("protein_input", False)
+        if protein_only:
+            logger.info("\t...skipping low-concern blastn and cmscan (protein input).")
+            self.reset_query_statuses(ScreenStep.LOW_CONCERN_DNA, ScreenStatus.SKIP)
+            self.reset_query_statuses(ScreenStep.LOW_CONCERN_RNA, ScreenStatus.SKIP)
+            rna_handler = None
+            dna_handler = None
```
manu-script, line 854: `ScreenStep.LOW_CONCERN_DNA` and `LOW_CONCERN_RNA` both map to the single shared `low_concern` field (`result.py:335-337`), and `parse_low_concern_hits` uses that same field as its per-query skip gate:

```python
if query.result.status.low_concern == ScreenStatus.SKIP:
    continue        # check_low_concern.py:546-550
```

So these two resets mark *every* query SKIP before the parser runs, and the low-concern hmmscan results computed just above are never applied and no hit is ever cleared on a `--protein` run. The HMMER step is the one low-concern check that does work on protein input, so this silently disables the whole step rather than just its blastn/cmscan halves.

Suggest dropping the resets and letting `parse_low_concern_hits` handle the `None` handlers (it already tolerates them and yields empty DNA/RNA frames):

```suggestion
        if protein_only:
            logger.info("\t...skipping low-concern blastn and cmscan (protein input).")
            rna_handler = None
            dna_handler = None
```

manu-script (commented):
Partial review of the protein-input path with three findings below. Still verifying a few others; will follow up.

commec/config/screen_io.py
```diff
@@ -128,38 +130,38 @@ def parse_input_fasta(self) -> dict[str, Query]:
...
-                        "Query %s: substituted %i non-IUPAC nucleotide characters with 'N'",
-                        record.id,
-                        substitutions,
-                    )
-                    # A high proportion of masked characters suggests the input isn't a
-                    # nucleotide sequence (e.g. a protein FASTA), in which case screening
-                    # results are meaningless
-                    proportion_substituted = substitutions / len(record.seq)
-                    if proportion_substituted > NON_IUPAC_SUBSTITUTION_THRESHOLD:
+                protein = self.config.get("protein_input", False)
```
manu-script, line 133: `_write_clean_fasta` replaces every non-ASCII sequence character with `_` to keep hit coordinates stable, and relies on `substitute_non_iupac` to turn those placeholders into `N` — see its own docstring at :309-311 ("The underscores then are replaced with 'N' by `substitute_non_iupac` masking"). Skipping the call for protein input leaves literal `_` characters in the sequence handed to hmmscan and blastp, and also skips the upper-casing `substitute_non_iupac` performed.

Protein input needs the equivalent masking against the amino acid alphabet rather than none — e.g. a `substitute_non_iupac_aa` over `IUPACData.extended_protein_letters` (the 20 standard residues plus the `BXZJUO` ambiguity/rare codes, which Biopython already bundles) plus `*` for translation stops, substituting `X` for anything else. The "may not be a nucleotide sequence" warning below deserves a protein counterpart too: as written, a nucleotide FASTA run with `--protein` passes without any complaint.

commec/screen.py
```diff
@@ -699,6 +729,32 @@ def screen_proteins(self):
...
             )
 
+        # Step 5: run blastp for protein input queries; skip if no protein sequences were found
+        protein_input_has_sequences = (
+            self.database_tools.regulated_protein_blastp is not None
+            and os.path.isfile(self.params.protein_path)
+            and os.path.getsize(self.params.protein_path) > 0
+        )
+        if protein_input_has_sequences:
+            self.database_tools.regulated_protein_blastp.search()
```
manu-script, line 739: `protein_search_info` is assigned unconditionally from `_tools.regulated_protein` (the blastx handler) at :543-545. On a `--protein` run the search that actually produced the taxonomy hits is blastp, so the report records the wrong tool version and database metadata.

Record the handler that actually ran — either key the assignment at :543 off `params.config["protein_input"]`, or set it here alongside the blastp search:

```python
self.screen_data.database_info.search_tool_info.protein_search_info = (
    self.database_tools.regulated_protein_blastp.get_version_information()
)
```

commec/screen.py
krishnanj: Thanks Manu for all the comments and feedback. Fixed in 7c658df. Removed both `reset_query_statuses` calls so `parse_low_concern_hits` sees the correct skip gate state and applies HMMER results on protein input. Added a regression test (`test_protein_query_low_concern_hmmer_clears_taxonomy_flag`) that places a TAXONOMY_AA flag and a LOW_CONCERN_PROTEIN hit on a protein query and asserts `CLEARED_FLAG`.

commec/config/screen_io.py
krishnanj: Fixed in 7c658df. Added `substitute_non_iupac_aa` using `IUPACData.extended_protein_letters`, substituting 'X' for any character outside that alphabet (including the `_` placeholder left by `_write_clean_fasta`). Upper-casing and a high-substitution-rate warning are both included, mirroring the nucleotide path. Five parametrized unit tests added to `test_screen_io.py`.

commec/screen.py
krishnanj: Fixed in 7c658df. `protein_search_info` now reads from `regulated_protein_blastp` when `protein_input` is True, and from `regulated_protein` (blastx) otherwise.

krishnanj:
Done in 7c658df. Added a protein checkbox as step 3 in the Colab form with an explanatory caption; passes `--protein` to `commec screen` when checked. Also extended the file upload to accept `.faa` files.

krishnanj:
Fixed the lint failure in c640139. CI uses ruff 0.15.x while my local ruff is 0.16.x; the two versions disagree on wrapping a long `pd.DataFrame` constructor call in `test_check_biorisk.py`. Applied the wrap manually to match what 0.15.x expects.

